PR #44, opened 2026-08-04
perf(deploy): measured startup/teardown wins + observability triage

AdrianLlopart:
Stacked on #43 — targets `fix/x86-compile-bytecode`, not `master`.

Deploy startup/teardown latency and observability triage. Every claim below is a
measurement taken on this host (RTX 4070 Laptop 8 GB, SO-101 on `/dev/ttyACM0`),
not an estimate.

## What changed

| Area | Before → After |
|---|---|
| Policy-dep probe (CLI preflight + reasoner palette seed) | **6.74 s → 0.49 s** per process, identical verdict |
| VLA framework import on `deploy sim` | ~7 s moved off the first goal, overlaps HAL configure |
| First inference tick (ACT so101-pen) | **330.4 ms → 14.9 ms** — was 10× over the 33.3 ms budget |
| VRAM freed on rSkill swap | **0 MiB → all of it** (768.2 → 0.0 measured) |
| Event Log info rows | **121/s → 0/s**; 200-slot ring stopped cycling every 1.65 s |
| `deadline_missed` WARN | **400 lines → 1** for 400 consecutive misses |
| Dashboard `e-stops` counter | **structurally pinned at 0 → works** |

## Why

Three cold-start fixes landed on #43 (bytecode precompile, GIL relief, import
pre-warm). They killed the catastrophic case, but nobody had measured the whole
critical path, and the unload side was largely unimplemented. Separately the
dashboard Event Log was unreadable: four 30 Hz span families sat at `info`, and
the structlog→OTel bridge was hard-pinned to DEBUG with no configuration surface.

## Notable findings

- **`find_spec` does not avoid the cost.** The obvious fix for the import tax was
  `find_spec` instead of `import_module`. Measured: **6.64 s vs 6.57 s** — it must
  import the parent, and `lerobot/policies/__init__.py` eagerly imports every
  family's config. The fix is to probe the *top-level* package only. The old
  docstring claimed "≈100 ms"; it was wrong by ~65×.
- **`empty_cache()` alone frees nothing.** Every VLA adapter's `close()` flushed
  the cache while still holding the policy. Measured: 768.2 MiB → 768.2 MiB.
  Dropping the reference first → 0.0 MiB.
- **Nine declared telemetry signals had no producer**, one with a visible
  consequence: the `e-stops` tile read 0 no matter how many e-stops fired.
- One planned item was **dropped as unnecessary**: the reasoner's "5 s first-tick
  delay" does not exist — `_on_prompt` already forces an immediate tick.

## Verification

4379 unit tests pass. Component-level measurements on the real GPU (real ACT
checkpoint, real CUDA allocator, real protobuf spans). Falsification tests
included where the ordering is the fix — e.g. the CUDA test asserts
`empty_cache()` alone does *not* move the counter, then that the release does.

**Not yet done:** live full-graph `deploy sim` / `deploy run` baseline.

## Review notes

- The e-stop counter fix is observability-only and additive — no safety
  threshold, check or behaviour altered — but it touches the e-stop callback, so
  it wants a safety-WG eye per CLAUDE.md §3.
- The **e-stop latch bug is deliberately NOT fixed here**: `store.py` clears
  `estopped` only on `safety.severity == "ok"`, which no emitter produces (the
  kernel emits `warn`/`violation`/`info`). Fixing it changes what the safety
  kernel reports — a genuine safety change. Recorded in `docs/reference/telemetry.md`.
- `OPENRAL_LOG_LEVEL` now defaults to INFO. Debug log *records* no longer reach
  the dashboard by default; span rows are banded separately, so the DEBUG chip
  still shows the per-tick stream.

https://claude.ai/code/session_01NsYPKeTw2Lc1N5X1uj4cFi


python/reasoner/src/openral_reasoner/tool_use.py
```diff
@@ -724,6 +796,12 @@ def _build_curated_model(entry: ReasonerModel) -> ToolUseClient:
     """Build the client for a registry model from its resolved properties."""
     api_key = os.environ.get(REASONER_API_KEY_ENV, "").strip() or None
     endpoint_override = os.environ.get(REASONER_ENDPOINT_ENV, "").strip() or None
+    # A named endpoint resolves the same way here as on the uncurated path;
+    # otherwise `ENDPOINT=ollama` would be handed to the SDK as a literal URL.
+    if endpoint_override is not None:
+        preset = _ENDPOINT_PRESETS.get(endpoint_override.lower())
+        if preset is not None:
+            endpoint_override = preset.url
```
AdrianLlopart: **Curated path takes only `preset.url` and drops the other four fields.**

`_build_curated_model` resolves a named endpoint to its URL, then builds the client entirely from the registry entry — so `dialect`, `auth_required`, `timeout_s` and `tool_choice` are discarded. Those four are the whole reason `_EndpointPreset` is a NamedTuple rather than a `dict[str, str]`.

Verified by running the builder in this worktree:

| env | result |
|---|---|
| `MODEL=claude-opus-4-8 ENDPOINT=ollama` | `AnthropicToolUseClient` at `http://localhost:11434/v1` — Anthropic dialect against an OpenAI-only server |
| `MODEL=gpt-5.5 ENDPOINT=huggingface` | `tool_choice="required"`, which this file's own preset comment says the HF router 400s on, and `timeout=10.0` instead of the preset's 60 s |

In each case the reasoner configures "successfully" and then emits no tool calls for the entire deploy — the worst failure shape, because bringup looks clean.

Either apply the preset properly on this path, or reject a named endpoint for a curated model with an explicit error. Silently honouring one field out of five is the one option that should not survive.

python/reasoner/src/openral_reasoner/tool_use.py
```diff
@@ -806,19 +884,32 @@ def _build_managed_local(
...
-    endpoint = os.environ.get(REASONER_ENDPOINT_ENV, "").strip()
+    raw_endpoint = os.environ.get(REASONER_ENDPOINT_ENV, "").strip()
     dialect = os.environ.get(REASONER_DIALECT_ENV, "").strip().lower()
-    if not endpoint or dialect not in {"anthropic", "openai"}:
+    preset = _ENDPOINT_PRESETS.get(raw_endpoint.lower())
+    if preset is not None:
+        endpoint, dialect = preset.url, dialect or preset.dialect
+    else:
+        endpoint = raw_endpoint or None
+    if endpoint is None or dialect not in {"anthropic", "openai"}:
```
AdrianLlopart, line 941: **The `anthropic` preset is unreachable.**

Its `url=None` means "use the Anthropic SDK's default host", but the guard on this line treats `endpoint is None` as "endpoint not configured" and raises. `anthropic` is the only preset with `url=None`, so it is the only one that is 100% dead.

```
$ MODEL=claude-haiku-4-5 ENDPOINT=anthropic API_KEY=sk-test
ROSConfigError: ... set OPENRAL_REASONER_ENDPOINT to a named endpoint
                 (['anthropic', 'deepseek', ...]) or to a URL plus OPENRAL_REASONER_DIALECT
```

The error instructs the operator to set exactly the variable they already set. `on_configure` catches `ROSConfigError` and returns FAILURE, so the deploy comes up with no reasoner at all.

Two sentinels are being multiplexed onto one `None`. `test_endpoint_presets.py` covers ollama / openrouter / huggingface / vllm / bare-URL but never `anthropic`, which is why the dead branch shipped.

python/observability/src/openral_observability/dashboard/store.py
```diff
@@ -651,17 +690,25 @@ def _record_span(self, span: Span) -> None:
...
+        spans took info *generation* to zero, but the rows an operator actually
+        wants — `deploy.bringup`, `rskill.execute` — still share one FIFO with
+        the debug stream, so they were evicted within seconds of being emitted.
+        Only the reasoner's ERROR bringup row survived, because errors were the
+        only thing mirrored here. An info row that cannot outlive the flood is
+        no more useful than one that was never emitted.
         """
         self._events.append(ev)
-        if ev.severity in _ERROR_SEVERITIES or ev.kind in _PROTECTED_EVENT_KINDS:
+        if ev.severity != "debug" or ev.kind in _PROTECTED_EVENT_KINDS:
```
AdrianLlopart: **This trades info-row eviction for safety-event eviction — the worse of the two.**

`_ERROR_EVENT_RING_SIZE = 64`. Before this change the protected lane held only errors plus `openral.event.skill_failure`. Now every non-debug row lands there, so routine INFO competes with safety events for the same 64 slots.

`world.scene_objects` alone fires at **~0.10/s** (measured on this branch), i.e. ~6 rows/min, so the lane fully cycles in roughly **11 minutes of an idle scene** — before counting `deploy.bringup`, `reasoner.tick` and `rskill.execute`. A `safety.violation` or `estop_requested` from minute 1 is gone by minute 12.

That is precisely the "counter goes up, no trace" failure the lane was added to prevent, and the docstring immediately above still describes the lane as the thing that keeps `estop` / `safety.violation` findable.

The two goals need two mechanisms — e.g. keep the lane for safety kinds and give headline INFO its own ring, or size/partition the lane by class — rather than one FIFO serving both.

python/observability/src/openral_observability/tracing.py
```diff
@@ -90,8 +108,21 @@ def inference_span(
...
-        yield span
+        try:
+            yield span
+        finally:
+            # Label set stays closed (design §9): `kind` only. Device/engine
+            # ride the span, where cardinality does not matter.
+            with contextlib.suppress(Exception):
+                from openral_observability import metrics as _metrics
+
+                _metrics.record_histogram_ms(
```
AdrianLlopart, line 146: **`openral.inference.duration` is now recorded twice per tick on the eval/sim path.**

This block was added so the deploy path would emit the histogram — it genuinely did not before, which the live SO-101 run confirmed. But `InferenceRunnerBase._record_tick_telemetry` (`python/runner/src/openral_runner/base.py:292`) already records the same instrument and was never de-wired.

Any path through `InferenceRunnerBase` (eval, sim, `openral benchmark run`) now produces two data points per chunk inference:

- `runner/base.py:292` with `base_attrs`
- `tracing.py:121` with `{kind: ...}`

Sample counts double and p95 is computed over a mixed population of two label sets measuring overlapping-but-different spans of work, so the budget threshold line on the dashboard reads against a distribution matching no single measurement.

The deploy path never instantiates the runner, which is exactly why the live verification did not catch this — the fix was validated only on the path that was previously silent.

python/observability/src/openral_observability/metrics.py
```diff
@@ -282,6 +290,13 @@ def get_observability_export_failures() -> Counter:
     """``openral.observability.export_failures`` — dropped OTLP batches.
 
     Labels: ``signal_kind`` (``trace`` | ``metric`` | ``log``).
+
+    .. warning::
```
AdrianLlopart: **Stale `.. warning::` — this PR wired the counter it says is unwired.**

`_sdk._FailureCountingSpanExporter` (added in `43449fd`) increments `openral.observability.export_failures`, but this docstring still reads:

> **Declared but never recorded.** Nothing increments it ... Wiring it means hooking the SDK's exporter failure path, not just adding an `add(1)`.

That instruction describes work the same PR completed. A maintainer reading it concludes the counter is inert and either deletes it or skips it while diagnosing dropped OTLP batches — the diagnosis it now supports.

Note the warning is accurate for `get_inference_timeouts()` a few functions up, which really is still unwired (tracked in #49) — so the two need to stop saying the same thing.

AdrianLlopart:
**Review finding — `cpp/openral_safety_kernel/README.md:80` (posted here: the file is not in this PR's diff, so it takes no inline comment).**

Line 80 still lists:

```
| `safety.clamped` | `false` (kernel rejects, never clamps) |
```

`43449fd` removed the attribute from `lifecycle_kernel.cpp` and `supervisor_node.py`, and `docs/reference/telemetry.md:236` records it as removed. The kernel README is the reference for the safety span's attribute set, so the two now contradict each other — a consumer written against this table parses for a key that is never emitted.

CLAUDE.md §1.14 requires docs to travel with the code in the same commit range; this one was missed while the Python-side docs were updated. It also bears on #47: a safety-WG reviewer checking the clamp deletion against its own documentation finds the docs disagreeing about what the kernel emits.

AdrianLlopart:
Per CLAUDE.md §3 and issue #47: this PR touches \`packages/openral_safety/\`, \`python/hal/src/openral_hal/lifecycle.py\`, and \`cpp/openral_safety_kernel/\` across three commits (\`bfedd5e\`, \`43449fd\` x2) and needs explicit safety-WG sign-off before merge.

Hazard-log entry opened: OpenRAL/management#6 (Entry 007) — independently re-derived hazard analysis for all three changes. The latch auto-clear in \`43449fd\` (store.py, e-stop indicator now clears on a clean pass from a real kernel instead of only via manual reset) is the one flagged for extra scrutiny.

Requesting a safety-WG reviewer be assigned to this PR.

PR #51, opened 2026-08-04
fix(rskill,observability): drop the unenforced ROSInferenceTimeout contract (issue #49)

### LINKED ISSUE #49: runner: ROSInferenceTimeout is declared but never raised — implement the timeout or delete the contract
## Summary

`ROSInferenceTimeout` is declared in the exception hierarchy (CLAUDE.md §5) and
documented as part of the inference contract, but **no backend raises it**. The
matching metric `openral.inference.timeouts` therefore has no producer.

The gap is not a missing `add(1)` — it is that the timeout itself does not
exist. Wiring the counter means implementing enforcement.

## Current state

`ROSInferenceTimeout` appears only in docstrings, as a contract no backend
enforces:

* `python/rskill/src/openral_rskill/runtime.py:73` — *"Raises
  `ROSInferenceTimeout`: If the backend exceeds its wall-clock budget."*
* `python/wam/src/openral_wam/protocol.py:70` — *"Raises `ROSInferenceTimeout`:
  When the model exceeds its declared …"*

`openral_observability.metrics.get_inference_timeouts()` carries an explicit
`.. warning::` saying so, and `docs/reference/telemetry.md` lists it under
"declared, not emitted". PR #44 audited the four such signals: the e-stop
counter got a producer, `openral.safety.clamps` was deleted as a concept that
does not exist, `observability.export_failures` got a producer — this one was
left open **deliberately**, because it needs a design decision rather than a
patch.

## Why it matters

Per-skill latency budgets are contractual (CLAUDE.md §3, "latency budget is
contractual"), and manifests declare them. Today a policy that blows its budget
is caught only downstream:

* `rskill_runner_node._deadline_lapsed` aborts the **goal** on the execution
  budget, between steps — a blocking `skill.step()` overruns it by up to one
  full step, and the abort is per-goal, not per-inference;
* the reasoner's patience ceiling cancels after N seconds of no progress.

Neither bounds a single `forward()`. A policy that hangs inside torch is
observable only as a missing chunk. That is the hole `ROSInferenceTimeout` was
declared to fill.

## Why this needs an ADR, not a PR

The hard part is enforcement semantics on the actuation path:

1. **Where does the clock live?** A wall-clock watchdog around `forward()`
   cannot interrupt CUDA work already queued; a cooperative check only fires
   between chunks, which is what the existing deadline check already does.
2. **What happens to the robot on a raise?** An inference timeout mid-execution
   leaves the HAL without a fresh chunk. Hold last? Ramp to zero velocity?
   Escalate to the safety kernel? This is a §3 safety-layer question, not a
   runner detail.
3. **Sidecar backends differ.** ZMQ sidecars (LingBot-VLA, RLDX-1,
   InternVLA-N1, the B1K GR00T runtime) already have a transport timeout with
   its own failure mode; in-process backends have none. Unifying them is the
   actual work.
4. **Budget source.** The manifest declares latency budgets, but per-chunk vs
   per-goal is not currently distinguished in the schema.

## Resolution paths

* **Implement it** — ADR covering (1)–(4), then the counter wires itself.
* **Delete it** — drop `ROSInferenceTimeout` from the hierarchy, the two
  docstrings, and the metric, as was done for the clamp concept in #44. Honest,
  and cheap, if per-goal deadline enforcement is judged sufficient.

Either is better than the current state, where the exception hierarchy
advertises a guarantee the runtime does not provide.

## Acceptance

- [ ] ADR recorded in `OpenRAL/management` deciding implement-vs-delete
- [ ] If implement: a backend raises it, the counter records it, and a test
      proves a hung `forward()` is bounded
- [ ] If delete: symbol, docstrings, metric and the telemetry-doc row all go
- [ ] `docs/reference/telemetry.md` "declared, not emitted" row resolved either way

AdrianLlopart:
## What changed

`ROSInferenceTimeout` and its producerless metric are deleted.

| Removed | Location |
|---|---|
| `class ROSInferenceTimeout(ROSRuntimeError)` | `python/core/src/openral_core/exceptions.py` (+ `__all__`) |
| `ROSInferenceTimeout` re-export | `python/core/src/openral_core/__init__.py` (import + `__all__`) |
| `get_inference_timeouts()` | `python/observability/src/openral_observability/metrics.py` (+ `__all__`) |
| `METRIC_INFERENCE_TIMEOUTS` (`openral.inference.timeouts`) | `python/observability/src/openral_observability/semconv.py` (+ `__all__`) |
| `Raises: ROSInferenceTimeout` on `Runtime.infer` | `python/rskill/src/openral_rskill/runtime.py` |
| `Raises: ROSInferenceTimeout` on `WorldModel.rollout` | `python/wam/src/openral_wam/protocol.py` |
| prose mention | `python/rskill/src/openral_rskill/testing.py` |
| inventory rows | `docs/methods/00-core-schemas.md`, `docs/methods/06-reasoning-wam-safety-observability.md` |
| exception + metrics card text | `docs/architecture/repo-state-map.html` |
| §5 example list entry | `CLAUDE.md` |

Plus a `### Removed` CHANGELOG entry under `[Unreleased]`, and `(LNN)` markers
refreshed via `tools/refresh_methods_linenos.py` (the only marker churn in
`docs/methods/` is the -1/-5 line shift caused by these deletions).

## Why

The owner chose the issue's **"Delete it"** path over **"Implement it"**.

`ROSInferenceTimeout` was declared in the exception hierarchy and promised by
two `Raises:` docstrings, but **no backend ever raised it** — so
`openral.inference.timeouts` had no producer. A contract nothing enforces reads
as a guarantee to anyone writing a backend or a dashboard panel; CLAUDE.md §1.2
(truth over plausibility) forbids leaving that in place. This is the fourth of
the four declared-not-emitted signals audited in #44, deliberately left open
there because it needed a decision rather than a patch.

Implementing it honestly is a much larger safety-layer design project: which
clock owns the budget, what the robot does when the exception fires
mid-forward-pass (a partially-consumed action chunk is not a safe state), how
in-process and ZMQ-sidecar backends unify under one deadline, and where the
budget is sourced from. That needs its own ADR and safety-WG review, and is not
in scope.

**No control is lost — it never was one.** The implemented deadline mechanisms
are untouched:

- `rskill_runner_node._run_until_done_or_deadline` — goal-level abort on the
  ROS action, driven by the goal's `deadline_s`, falling back to the manifest
  latency budget when `deadline_s<=0`.
- The reasoner patience ceiling (`resolve_patience_s`) + bounded replanning
  ladder.
- `openral.tick.budget_violations` / `openral.tick.deadline_misses` — the
  tick-cadence counters that *do* have producers.
- `openral_rskill.testing.assert_within_budget` — CI-time manifest budget
  assertion.

Per-inference (as opposed to per-goal) enforcement stays available to reopen as
its own ADR whenever a concrete hazard motivates it.

**Not a safety change.** No actuation path, no `packages/openral_safety/`, no
`cpp/openral_safety_kernel/` file is touched.

## How tested

Run against the main venv with the worktree source dirs prepended to
`PYTHONPATH` (ROS paths preserved).

- `grep -rn ROSInferenceTimeout` → **zero** hits repo-wide.
- `grep -rn 'inference_timeouts|inference\.timeouts|INFERENCE_TIMEOUTS'` → one
  hit, in `docs/audit/observability-traceability-report.html`. **Left
  deliberately**: that file is a dated point-in-time audit snapshot (header
  reads *"Branch: … · Date: 2026-06-21"*, and its `semconv.py:211–236` citation
  is already stale). Editing it would falsify a historical record rather than
  fix a live inventory.
- Import check: `openral_core`, `openral_rskill.runtime`, `openral_wam.protocol`,
  `openral_observability.metrics` all import clean; the four deleted names are
  gone from their modules and `__all__`s.
- `pytest tests/unit` → **4168 passed, 38 skipped, 11 failed**. All 11 failures
  reproduce identically on a pristine `origin/master` tree (verified by
  stashing this diff and re-running): worktree-env artifacts in
  `test_cli_deploy_sim.py`, `test_deps_install_plans.py`,
  `test_rldx_adapter_auto_spawn.py`, `test_sensor_ros_publisher.py`. None
  reference inference timeouts.
- `ruff check python/ docs/` → clean; `ruff format --check` on all 7 touched
  Python files → already formatted.
- `mypy --strict` per package → zero errors in any touched file (remaining
  errors are pre-existing missing-stub noise in `rclpy` / `lerobot` /
  `gym_aloha`).
- `pytest --doctest-modules` on the three touched docstring-bearing modules → 3
  passed.
- `tools/refresh_methods_linenos.py --check` → clean.

## ADR

Recorded as a 2026-08-04 amendment to **ADR-0010 (Inference runner)** in the
private `OpenRAL/management` repo, on local branch
`adr/0010-drop-inference-timeout-contract` (unpushed, awaiting owner review).

Fixes #49

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


AdrianLlopart:
Rebased onto the new base (`refactor/deployment_optimization`) and resolved the two conflicts:

- **`metrics.py`**: kept the base's clamp-concept deletion AND this PR's `get_inference_timeouts` deletion (the naive resolution would have resurrected `get_safety_clamps` against a `semconv` that no longer defines its constant — import error).
- **`docs/methods/06`**: both rows gone; `(LNN)` markers refreshed tree-wide (`--check` clean).
- The base's rewritten `test_declared_not_emitted.py` still expected `get_inference_timeouts` in its unrecorded set — adapted in `167eb18` (set emptied, both `telemetry.md` rows dropped, count corrected to six). `python/observability/tests`: **220 passed**.

The robot-manifest floor-fix cherry-pick dropped out of the branch automatically (already in this base via `fix/x86-compile-bytecode`).

⚠️ CI note: with a non-`master` base, `quality` and `select-and-test` no longer trigger on this PR (the branch-filter gap tracked in #54) — only DCO + docker-build run. The suites were run locally instead (documented per-commit).

AdrianLlopart:
Folded into #43 as `4abd82a`.

The stack was consolidated into `fix/x86-compile-bytecode` in dependency order — #44 (the parent) first, then #51, #52, #53 — so this branch's commits are now contained in #43 and it has nothing left to merge on its own. Verified with `git merge-base --is-ancestor`.

Closing rather than retargeting so the review history here stays intact and readable. Nothing is lost: reopen if you disagree with a conflict resolution — they are all itemised, with the reasoning, in `4abd82a`'s commit message.

AdrianLlopart:
**Correction to the SHA above:** the branch was rewritten after that comment to add DCO sign-offs to five pre-existing commits inherited from #44, so the merge commit for this PR is now `87aa6e3` (was `4abd82a`).

The rewrite was `git filter-branch --msg-filter` — **commit messages only**. Trees are unchanged by construction, and this was verified: the branch tree hash is identical before and after (`078209a0`), `git diff` between the two is empty, and the commit/merge counts are unchanged (71 / 5). No conflict resolution moved.

AdrianLlopart:
**Correction to the SHA above:** the branch was rewritten after that comment to add DCO sign-offs to five pre-existing commits inherited from #44, so the merge commit for this PR is now `87aa6e3` (was `4abd82a`).

The rewrite was `git filter-branch --msg-filter` — **commit messages only**. Trees are unchanged by construction, and this was verified: the branch tree hash is identical before and after (`078209a0`), `git diff` between the two is empty, and the commit/merge counts are unchanged (71 / 5). No conflict resolution moved.

PR #52, opened 2026-08-04
ci: wire OPENRAL_TEST_ROS_LIVE tests into CI (issue #46)

### LINKED ISSUE #46: ci: 33 OPENRAL_TEST_ROS_LIVE tests run in no workflow and no Justfile recipe
## Summary

**33 tests across 9 files are gated on `OPENRAL_TEST_ROS_LIVE=1`, and that
variable appears in no GitHub workflow and no `Justfile` recipe.** They run
only when a developer sets the variable by hand. Nothing in CI, `just test`, or
`just lint` ever executes them.

```
$ grep -rn OPENRAL_TEST_ROS_LIVE .github/ Justfile | wc -l
0
```

## Why this matters — a concrete rot

PR #44 found a VRAM pre-dispatch gate (`_refuse_unfittable_vla`) that was
**structurally dead for every VLA in the palette**. `_manifest_for_rskill`
resolved ids against the install registry
(`~/.local/share/openral/rskills.json`), which a `rskill_search_paths`-seeded
palette never populates — 48 palette manifests vs 3 registry entries, none of
them VLAs — so the gate early-returned `False` and let every dispatch through.
It surfaced only on real hardware, as a 20 s patience-ceiling timeout on a skill
the CLI preflight had already declared unfittable.

Its one test, `tests/integration/test_reasoner_vram_pair_refusal.py`, is in this
gated set — and would not have caught it anyway (it stubbed the broken method;
fixed in #44). But the CI gap is the reason a gated safety-adjacent gate can rot
for as long as it likes without anyone noticing.

## The gated files

| file | tests |
|---|---|
| `packages/openral_reasoner_ros/test/test_reasoner_node.py` | |
| `tests/integration/test_reasoner_node_end_to_end.py` | |
| `tests/integration/test_reasoner_dispatch_robustness.py` | |
| `tests/integration/test_reasoner_async_llm.py` | |
| `tests/integration/test_reasoner_vram_pair_refusal.py` | |
| `tests/integration/test_reasoner_vram_eviction.py` | |
| `tests/integration/test_reasoner_palette_primes_vram_gate.py` | |
| `tests/integration/test_critic_producer_node.py` | |
| `tests/unit/test_gstreamer_perception_tee.py` | |

33 gated tests total.

## Options

1. **Run them in the existing ROS CI job.** The workflows already build the
   colcon overlay for `launch_testing` integration tests, so the environment
   exists; this is a matter of exporting the variable for that step. Cheapest
   fix if the tests are stable there.
2. **A `just test-ros-live` recipe.** Does not close the CI gap, but makes the
   suite discoverable and runnable in one command instead of a docstring
   incantation. Worth doing regardless of (1).
3. **Retire the gate where it is no longer needed.** The variable predates some
   of the collection guards; some of these may run fine under the standard
   `pytest.importorskip("rclpy")` path that the rest of the suite uses, in which
   case the extra opt-in is pure friction.

Whichever route, the acceptance criterion is the same: **a deliberately broken
`_refuse_unfittable_vla` must turn CI red.** That is testable today —
`tests/integration/test_reasoner_palette_primes_vram_gate.py` was falsified
against the unfixed code in #44 and fails cleanly.

## Not in scope

HIL tests (`tests/hil/`), which are correctly gated on self-hosted lab runners
per CLAUDE.md §2.

AdrianLlopart:
## What changed

**The issue's premise needed correcting first.** It says "the workflows already
build the colcon overlay for `launch_testing` integration tests, so the
environment exists". They do not — there is no ROS CI job:

- `test-selective.yml` is the only test workflow with a PR trigger. It runs on a
  plain `ubuntu-24.04` runner with **no ROS 2 at all**; every gated test
  `importorskip`s there even with the variable set.
- `test-python.yml` and `hal.yml` are `workflow_dispatch:`-only ("out of Actions
  credits").
- `tools/test_selection.toml` still lists `.github/workflows/test-ros2.yml` in
  `full_run_globs` — that workflow does not exist. (Left alone here; inert
  dangling glob, pre-existing, not this PR's scope per CLAUDE.md §1.15.)

The one PR-triggered job with a real ROS 2 + colcon overlay is **`docker-build`**,
which builds `openral:x86` (ROS Jazzy + `colcon build` of `openral_msgs`,
`openral_reasoner_ros`, `openral_prompt_router`, …). That is where the tests now run.

1. **`scripts/ros_live_tests.sh`** — new. Owns the target list and sets
   `OPENRAL_TEST_ROS_LIVE=1`. Single source of truth: both callers exec it, so
   the dev loop and CI cannot drift apart.
2. **`.github/workflows/docker-build.yml`** — new "Live ROS tests" step runs that
   script inside the freshly-built image. `tests/`, `conftest.py` and `scripts/`
   are bind-mounted (the deploy image ships a runtime, not a test tree), which
   also guarantees the PR's tests run against the PR's image. `paths:` extended
   so a diff touching only these tests still triggers the job.
3. **`Justfile`** — `just test-ros-live` (issue option 2), same script.
4. **Docs** — `docs/contributing/development.md` gains a "live-ROS suite"
   section; the six test-module docstrings swap their hand-typed
   `OPENRAL_TEST_ROS_LIVE=1 uv run pytest …` incantations for the one command.

### File-by-file decision (option 1 vs 3), as requested

| File | Decision | Why |
| --- | --- | --- |
| `test_reasoner_node_end_to_end.py` (21) | → CI | needs real rclpy + DDS + overlay; all present in `openral:x86` |
| `test_reasoner_dispatch_robustness.py` (3) | → CI | same |
| `test_reasoner_async_llm.py` (5) | → CI | same |
| `test_reasoner_vram_pair_refusal.py` (2) | → CI | same; **this is the falsification test** |
| `test_reasoner_vram_eviction.py` (1) | → CI | same |
| `test_critic_producer_node.py` (1) | → CI | same |
| `tests/unit/test_gstreamer_perception_tee.py` (1 gated) | **gate kept** | also needs PyGObject. `openral:x86` is *the* GStreamer-free image by design (media stack is OpenRAL Pro), so no CI surface has `gi` + `rclpy` together — wiring it would buy a guaranteed skip. Option 3 rejected too: the gate is a documented **crash guard** (rclpy/DDS vs. a glib pulled in by torch/pyarrow), not just an availability check. Rationale recorded in-file. |
| `packages/openral_reasoner_ros/test/test_reasoner_node.py` | **no change needed** | it is **not gated** — plain `pytest.importorskip`; it only *mentioned* the variable in a stale docstring, which is why the issue's grep caught it. Docstring corrected. |

Test counts: 21+3+5+2+1+1 = **33**, matching the issue exactly. The gstreamer
file's 1 gated test and the un-gated colcon shim are the other two of the nine.

`tests/integration/test_reasoner_palette_primes_vram_gate.py` is listed in the
issue but is **not on `master`** — it lands with PR #44 (still open, based on
`fix/x86-compile-bytecode`). It is not in `TARGETS`; add it there when #44 merges.

## Why

A gated safety-adjacent gate can rot indefinitely with nothing watching. PR #44
found `_refuse_unfittable_vla` structurally dead for every VLA in the palette;
it surfaced only on real hardware, as a 20 s patience-ceiling timeout.

## How tested

The local host has ROS 2 Jazzy and a prebuilt `openral:x86`, so **the exact CI
command was run locally**, not approximated.

**Green** — the literal `docker run …` line from the new workflow step:

```
33 passed in 39.82s      EXIT=0        (~40 s, no GPU on the container)
```

**Red (acceptance criterion)** — `_refuse_unfittable_vla` made structurally dead
(`return False` inserted before its first statement, reproducing the PR #44 bug),
mounted over the image's installed `reasoner_node.py`, same command:

```
3 failed, 30 passed in 61.87s        DOCKER_EXIT=1

FAILED test_reasoner_dispatch_robustness.py::test_resident_vla_redispatch_skips_the_free_vram_probe
FAILED test_reasoner_vram_pair_refusal.py::test_execute_rskill_refused_when_vla_reward_pair_exceeds_vram
FAILED test_reasoner_vram_pair_refusal.py::test_execute_rskill_refused_when_live_free_vram_is_below_vla_min
```

The deliberate breakage was never committed — it lived only in a scratch file
bind-mounted over the container path. Worth noting: on `master` the pair-refusal
test does **not** merely stub the broken method (as the issue suspected); it
catches the dead gate on its own, so the criterion is met without #44's palette test.

Also run: `bash -n scripts/ros_live_tests.sh`, `just --list`, workflow YAML
parse, `ruff check` + `ruff format --check` clean on all changed Python.

**Not verifiable locally:** the real GitHub Actions run — whether the ~16 GB
image build plus this step fits the hosted runner's time/disk budget. The step
itself adds ~40 s and no disk. Only the first CI run on this PR can confirm that,
and this PR touches `.github/workflows/docker-build.yml`, so it will trigger.

Fixes #46

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -7,8 +7,9 @@ name: docker-build
 # weeks):
 #   * pull_request → build + smoke only (no push)
 #   * push to master → build + push openral:x86 to GHCR
-# The `paths:` filter means it only runs when the image's build inputs change,
-# not on every commit. workflow_dispatch stays for manual runs (no push).
+# The `paths:` filter means it only runs when the image's build inputs change
+# (or the live-ROS tests it hosts), not on every commit. workflow_dispatch stays
+# for manual runs (no push).
 on:
   workflow_dispatch:
   pull_request:
@@ -20,6 +21,11 @@ on:
       - cpp/**
       - python/**
       - .github/workflows/docker-build.yml
+      # The live-ROS suite runs in this job and nowhere else (issue #46), so a
+      # diff that only touches those tests must still rebuild + run them.
+      - scripts/ros_live_tests.sh
+      - tests/integration/test_reasoner_*.py
+      - tests/integration/test_critic_producer_node.py
   push:
     branches: [master]
     paths:
@@ -30,6 +36,11 @@ on:
       - cpp/**
       - python/**
       - .github/workflows/docker-build.yml
+      # The live-ROS suite runs in this job and nowhere else (issue #46), so a
+      # diff that only touches those tests must still rebuild + run them.
+      - scripts/ros_live_tests.sh
+      - tests/integration/test_reasoner_*.py
+      - tests/integration/test_critic_producer_node.py
 
 permissions:
   contents: read
@@ -77,6 +88,31 @@ jobs:
             openral deploy run --config scenes/deploy/so101_bench.yaml --dry-run | grep -q "hal_mode:=real"
             echo "OK: deploy run --dry-run resolves"
           '
+
+      # Live-ROS suite (issue #46). These tests are gated on
+      # OPENRAL_TEST_ROS_LIVE=1 because they need a real rclpy, a real DDS
+      # graph, and the colcon-built openral_msgs / openral_reasoner_ros /
+      # openral_prompt_router overlay. This image is the ONLY CI surface that
+      # has any of that — the ubuntu-hosted test-selective runner has no ROS 2
+      # at all, so every one of them importorskips there. Without this step
+      # they run nowhere, and safety-adjacent gates like the reasoner's
+      # pre-dispatch VRAM refusal rot unnoticed (PR #44).
+      #
+      # tests/ + conftest.py + scripts/ are NOT baked into the deploy image
+      # (it ships a runtime, not a test tree), so bind-mount them from the
+      # checkout — that also guarantees the PR's version of the tests runs
+      # against the PR's image. No GPU on the runner and none needed: the
+      # free-VRAM readings are pinned at the nvidia-smi process boundary.
+      - name: Live ROS tests (OPENRAL_TEST_ROS_LIVE)
+        run: |
+          # --entrypoint /entrypoint.sh drops the image's hardcoded `openral`
+          # argument while keeping the ROS + colcon-overlay sourcing the tests
+          # depend on.
+          docker run --rm --entrypoint /entrypoint.sh \
+            -v "$PWD/tests:/workspace/tests:ro" \
+            -v "$PWD/conftest.py:/workspace/conftest.py:ro" \
+            -v "$PWD/scripts:/workspace/scripts:ro" \
+            openral:x86 bash scripts/ros_live_tests.sh -q
 
       # Default-entrypoint smoke: the steps above override --entrypoint, so
       # they never exercise what `docker run openral:x86 …` actually does.
--- a/Justfile
+++ b/Justfile
@@ -306,6 +306,15 @@ hal-twin-sweep:
 test-integration:
     PYTHONPATH="$(pwd)/packages/world_state:${PYTHONPATH}" uv run pytest tests/integration/ -v --tb=short
 
+# Real rclpy + a real DDS graph + the colcon overlay; no GPU needed. Requires
+#     source /opt/ros/jazzy/setup.bash && just ros2-build && source install/setup.bash
+# scripts/ros_live_tests.sh owns the target list, and the `docker-build`
+# workflow runs that same script inside `openral:x86` — so this recipe and CI
+# can never select a different set of tests. `-k <expr>` narrows the run.
+# Live-ROS suite (OPENRAL_TEST_ROS_LIVE=1) — see docs/contributing/development.md
+test-ros-live *args:
+    uv run bash scripts/ros_live_tests.sh -q {{args}}
+
 # Subset by keyword
 # `-p no:launch_testing -p no:launch_ros`: same ROS-env workaround as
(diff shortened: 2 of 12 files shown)


AdrianLlopart:
Folded into #43 as `4a1199e`.

The stack was consolidated into `fix/x86-compile-bytecode` in dependency order — #44 (the parent) first, then #51, #52, #53 — so this branch's commits are now contained in #43 and it has nothing left to merge on its own. Verified with `git merge-base --is-ancestor`.

Closing rather than retargeting so the review history here stays intact and readable. Nothing is lost: reopen if you disagree with a conflict resolution — they are all itemised, with the reasoning, in `4a1199e`'s commit message.

AdrianLlopart:
**Correction to the SHA above:** the branch was rewritten after that comment to add DCO sign-offs to five pre-existing commits inherited from #44, so the merge commit for this PR is now `e4b8daa` (was `4a1199e`).

The rewrite was `git filter-branch --msg-filter` — **commit messages only**. Trees are unchanged by construction, and this was verified: the branch tree hash is identical before and after (`078209a0`), `git diff` between the two is empty, and the commit/merge counts are unchanged (71 / 5). No conflict resolution moved.

AdrianLlopart:
**Correction to the SHA above:** the branch was rewritten after that comment to add DCO sign-offs to five pre-existing commits inherited from #44, so the merge commit for this PR is now `e4b8daa` (was `4a1199e`).

The rewrite was `git filter-branch --msg-filter` — **commit messages only**. Trees are unchanged by construction, and this was verified: the branch tree hash is identical before and after (`078209a0`), `git diff` between the two is empty, and the commit/merge counts are unchanged (71 / 5). No conflict resolution moved.

PR #80, opened 2026-08-08
fix(cli): ship ROS bootstrap in the wheel + detect unified-memory GPUs (GB10)

AdrianLlopart:
## Summary
- `openral install ros` no longer requires a repo clone: `scripts/bootstrap_ubuntu.sh`/`bootstrap_macos.sh` now ship inside the `openral-cli` wheel/sdist and are resolved via `importlib.resources`; `just bootstrap`, `scripts/install.sh`, and docs updated to match.
- Tier-0 installer now also installs `just` (user-local via `uv tool install rust-just`, non-fatal) so a freshly-curled host has it before cloning.
- GPU detection no longer reports unified-memory GPUs (GB10 / DGX Spark) as absent: NVML/`nvidia-smi` per-device probing degrades independently instead of aborting on `NVMLError_NotSupported(3)` for VRAM, with a system-RAM fallback for unified-memory pools; `nvcc` resolution now prefers `/usr/local/cuda/bin/nvcc` over a shadowing `/usr/bin/nvcc`.

## Test plan
- [x] `uv build python/cli` — bootstrap scripts present in both wheel and sdist, byte-identical to repo copy, `bash -n` clean
- [x] Real `openral install ros` run from an empty cwd with no `OPENRAL_REPO_ROOT`: locates the packaged script and executes it (fails only at the refused-sudo step, as expected)
- [x] New `TestPackagedBootstrapScripts` and `TestUnifiedMemoryHosts` (3 cases) fail pre-fix, pass post-fix
- [x] Live verification on a real DGX Spark (GB10, driver 580.126.09) over SSH: `openral doctor` now reports `GPU 0 | ok | NVIDIA GB10 (124610 MiB, sm_121)` and `cuMotion | ok | supported`
- [x] `tests/unit`: 4360 passed (4 worktree-environment-only failures, pass in main checkout)
- [x] ruff, ruff-format, `mypy --strict` on the touched packages, `tools/refresh_methods_linenos.py --check` all clean

https://claude.ai/code/session_012H3HyYGrRpHBgjbXZKpkQt


AdrianLlopart:
Superseded — this branch already landed as #73 (`4f1275d1`). Every file here is byte-identical to `master`, so this PR is a duplicate with an effectively empty diff. Closing and deleting the branch.

PR #101, opened 2026-08-10
fix(safety): scale normalized Cartesian predictions

AdrianLlopart:
## What changed

- Carry typed normalized Cartesian scales beside unchanged controller-space actions; predictive safety uses `clip(raw, -1, 1) * scale`.
- Clip explicitly declared normalized joint-velocity slots before safety/HAL dispatch.
- Replace circumscribed voxel spheres with exact occupied cubes.
- Commit multi-surface policy slots atomically and block the producer until `/openral/action_applied` confirms application.
- Preserve XR-1 RoboCasa365's trained 256x256 policy input while retaining higher-resolution source frames for perception.
- Remove the temporary XR-1/RoboCasa 5 Hz scene cap; the normal runner/hardware path sets the ceiling.
- Keep deploy-sim environments persistent: deploy ignores simulator success/terminal metadata and never resets after startup. `openral sim run` remains the evaluator and owns episode resets.
- Require immutable revisions for every shipped custom-code runtime: XR-1, OpenVLA, and LocateAnything.
- Correct PandaMobile collision FK and add deterministic startup seed support.

## Why

Direct XR-1 RoboCasa365 exposed several deploy integration mismatches: normalized controller inputs were interpreted as physical units, multi-surface actions were stepped independently/dropped under load, and checkpoint image resolution changed policy behavior. Follow-up review also found deploy-specific simulator-oracle completion, runtime episode resets, and a scene-specific 5 Hz workaround that do not belong in a deployment twin.

## Validation

- 4,501 unit tests passed
- all four C++ safety-kernel test binaries passed
- lint + strict mypy, schema drift, rSkill metadata drift, methods inventory, and strict docs build pass
- custom-code manifest audit covers XR-1, OpenVLA, and LocateAnything immutable revisions
- contract audit: normalized Cartesian scales apply to shipped RoboSuite/LIBERO delta policies; normalized joint bounds apply to XR-1/RLDX RoboCasa365 base slots; atomic application covers all multi-slot contracts; exact voxel geometry is global; exact 256px resize is currently evidenced only for XR-1 RoboCasa365
- pre-oracle-removal action-path video: https://github.com/user-attachments/assets/4fcfbf01-0df5-478d-84cc-ddcb3ceb5a82

## Safety posture

**Draft / do not merge yet.** The dense-kitchen OctoMap/Nav2 issue remains separate and unresolved; tracked in #102. This PR does not lower the world-voxel margin or whitelist an unlabeled voxel.

[...]

PR #103, opened 2026-08-10
fix(robot): align PandaMobile voxel collision model

AdrianLlopart:
## What changed

- Restore PandaMobile collision FK to the ground-level `mobilebase0_base` frame: 0.7 m pedestal + 0.333 m Franka joint-1 offset.
- Replace the seven hand-authored Panda arm capsules with OBBs derived from the compiled RoboCasa collision meshes.
- Add real RoboCasa regressions proving the FK root matches the simulator and every OBB encloses its mesh.
- Keep the configured 2 cm world-voxel margin, voxel data, and E-stop behavior unchanged.

## Why

PR #101 moved joint 1 from 1.033 m to 0.333 m after treating `base_link` as the elevated arm mount. The voxel grid is expressed relative to ground-level `mobilebase0_base`, placing every kernel link exactly 0.7 m too low and false-stopping on cabinet-door voxel 39263.

After correcting FK, full runs exposed additional clearance over-report from the rounded arm capsules. The OBBs preserve the complete real meshes while removing that approximation.

## How tested

- 11 targeted collision contract tests passed.
- Real RoboCasa FK/mesh regression passed for all seven arm links.
- Full 27-package ROS graph rebuilt from the stacked branch, including generated IDL and the C++ safety kernel.
- Clean zero Cartesian probe with OctoMap gate enabled: 1 safe chunk, 0 E-stops, 1,084 occupied voxels.
- One seed-1 run applied 208 XR-1 chunks and visibly placed the baguette in the cabinet: `/home/allopart/openral-runs/issue102_wrist_fix_xr1.mp4`.

## Remaining blocker

Do not merge yet. The Cartesian predictor adds 1 cm uncertainty to the configured 2 cm voxel margin even for a one-step action. Repeat runs can therefore stop at 2.5-3.0 cm mesh-OBB clearance. This requires Safety-WG calibration; the margin was not lowered or bypassed.

The world-state `staleness_latched` warnings are camera diagnostics: idle 512 px rendering is about 0.6 Hz while the fixed wall-clock threshold is 0.5 s. Joint state and voxels remain fresh.

## Stack

Base: #101 (`fix/cartesian-delta-safety-scale`). Merge #101 first, then this PR only after the human Safety-WG gate and remaining Nav2/full-trajectory evidence.


AdrianLlopart:
Final calibrated rerun: cameras remained fresh; simulated Cartesian first-step inflation was removed; sim voxel margin was reduced to exact overlap (0 m); transient occupancy requires two hits. XR-1 was dispatched automatically and reached 191 applied chunks, then exact OBB-vs-voxel overlap at tick 192 (`panda_link7` vs `voxel_63108`, `-0.000188 m`) E-stopped. Further margin reduction is impossible without disabling collision detection. Remaining blocker is attached-object semantics: remove the grasped object from the world OctoMap while attaching its payload geometry to the robot collision model. PR remains draft.

Issue #102, opened 2026-08-10
fix(sim): resolve PandaMobile floating voxels and Nav2 collision semantics


AdrianLlopart:
## Problem

RoboCasa `deploy sim` had an unresolved dense/floating voxel interaction around PandaMobile and Nav2.

> **Rewritten 2026-09-05.** This body described a reproduction that **no longer reproduces**, and half of it had already moved to #108 by comment while staying in the "Expected" list here. Nothing is silently dropped: every struck item below says what closed it.
>
> **Re-verified 2026-09-07 against every merged PR through `78fa614` (#239).** All four acceptance items are now closed. The gate was redefined by maintainer ruling to drop the S2 reasoner leg (the navigation path is proven independently on #108), and on that definition it is met by recorded evidence — see the comment below. Hazard-log Entry 022 is **ACCEPTED** ([`OpenRAL/management#35`](https://github.com/OpenRAL/management/pull/35)).

**Scope.** Narrowed by the scoping comment below to the **kernel half** — FK, voxel false-positives, attached-object semantics. The Nav2/costmap half is #108 and is not tracked here.

### The original reproduction, and what happened to it

- ~~With the OctoMap kernel gate enabled the valid nominal seed-1 start is rejected before motion: `panda_link4` vs `voxel_39263`, centre ≈ `[0.175, 0.075, 0.725]` in `base_link`, ≈ `-0.0526 m`.~~ **Does not reproduce.** `docs/reference/robocasa-start-state-census.md` measured every start state in the four-scene matrix — 4 scenes × 60 layouts = 240 states, each at its own seed, zero actions applied — and found **72 kernel stops with `UNEXPLAINED: 0`**. `panda_link4` dominates **0 of the 72**, despite having the worst corner slop in the model (88.22 mm); the stops are `panda_link2` (46) and `panda_link1` (14), 83.3% between them. 91.7% of all stops are envelope conservatism — the link is measurably clear and the OBB corner slop plus the 21.65 mm voxel inflation exceed the clearance — and at most 8.3% involve any contact at all. The `link4` class was removed by the stack that landed in between: #103 (base-relative frame), #104 (attached collision objects), #178 (world grid on the map's own lattice), #180 (intangible geometry filtered out of the depth cast).
[...]

PR #104, opened 2026-08-11
feat(safety): model attached collision objects

AdrianLlopart:
## What changed

- Add `openral_core.AttachedCollisionObject` and `AttachmentEvidenceKind` to the published `openral-core` contract.
- Add `openral_msgs/msg/AttachedCollisionObject` and carry multiple attachments through `WorldStateStamped`.
- Add atomic multi-object attachment storage to `WorldStateAggregator`.
- Round-trip sphere, capsule, and box payloads with optional mass/CoG/inertia and evidence provenance.
- Export JSON schemas and update methods/state-map documentation.

## Why

A grasped object must move from world occupancy into collision-active robot geometry. Gripper filtering or margin reduction cannot safely solve double-counting because the carried payload must still collide with the robot and environment.

Private decision: ADR-0092. Parent stack: #103.

## How tested

- 73 schema, fuzz, collision-contract, and ROS round-trip tests passed.
- Multiple attachments round-trip over real generated ROS IDL.
- `mypy --strict -p openral_core -p openral_world_state` passes.
- ROS overlay build passes for `openral_msgs` and `openral_world_state`.
- Full ruff/format/schema/method checks pass.

## Remaining WIP

- atomic attach/detach command surface and evidence producers;
- instance-aware RoboCasa/depth masking for multiple attached objects;
- bounded attached-payload ingestion and collision checking in the C++ kernel;
- multi-object deploy-sim grasp/carry/release tests.

## Safety posture

Draft only. No collision check is disabled by this slice; it adds state propagation but does not yet alter actuation.


AdrianLlopart:
Implementation update: added atomic `/openral/attachment_state` complete snapshots, exact MuJoCo body/descendant resolution in `SimAttachedHAL`, and attached-body union with the depth/OctoMap self-filter. A real seed-1 RoboCasa deploy-sim test attaches `obj_main` and `distr_counter_main` simultaneously, proves both leave the depth world, detaches one without reset, then restores the baseline. Sphere/capsule/box ROS round-trip and multi-object WorldState tests also pass. Kernel payload collision is being implemented in a parallel worktree.

AdrianLlopart:
End-to-end update: integrated fixed-capacity C++ payload collision (97 tests), producer-owned freshness, graph-clock alignment, and kernel revision acknowledgements. Sim masking now follows safe ordering: detach unmask first; additions mask only after `/openral/attachment_state_applied`. Live seed-1 RoboCasa validation: empty heartbeat acknowledged + safe action passed; two simultaneous objects acknowledged, occupied voxels fell 1084→1066, safe action passed; acknowledged oversized payload then E-stopped. Real deployment remains disabled pending a hardware evidence/payload manager. Automatic grasp/release evidence is the remaining WIP before XR-1 can use this without an external AttachmentState producer.

AdrianLlopart:
XR-1 was dispatched automatically on the attachment-enabled graph. It reached tick 144, then `panda_link5` overlapped voxel 61578. `WorldState.attached_objects` remained empty for the entire run, which confirms the new consumer/kernel path is armed but no automatic grasp producer has yet emitted the task object. The next implementation slice is the general evidence producer: exact MuJoCo contact/body identity in sim and gripper-force + perception-track confirmation on hardware.

AdrianLlopart:
Exact collision trace result: voxel 63179 came from MuJoCo `obj_g10` / body `obj_main` (the baguette). The real `robot0_link7_collision` mesh was 65.1 mm clear. MuJoCo contacts existed only between the baguette and gripper hand/fingers/pads (intended grasp contact, <0.8 mm penetration); no arm-link contact existed. The nearest current accepted depth hit was 57.5 mm from the occupied voxel center, with no raw hit within 5 cm. Root cause is the grasped baguette remaining in OctoMap plus voxel discretization/persistence expanding it into the wrist OBB—not physical arm collision and not the current mesh-derived OBB size.

AdrianLlopart:
Phase 1 gate passed. Attached objects now carry bounded multi-primitive geometry end-to-end in openral-core, openral-msgs, WorldState, sim masking, and the allocation-free C++ kernel. Validation: 105 kernel tests; 42 schema/unit tests; sourced ROS round-trip; two real RoboCasa tasks (`PickPlaceCounterToCabinet`, `PickPlaceCounterToSink`) each attached two unknown free bodies with two primitives/object, masked both, then detached/restored without reset. Phase 2 automatic simulator evidence is now unblocked; later phases remain gated.

AdrianLlopart:
## Local validation @ 50b57732 (RTX 4070 8 GB, deploy-sim, kernel gate + Nav2 + SLAM on, XR-1, seed 1 baguette)

**Q1 — automatic attachment evidence: PASS.** Both actuated runs: grasp → `attached_objects` populated automatically (`sim:obj_main`, attach_link `panda_link7`, `sim_contact` evidence, 16 box primitives, mass 0.057 kg), revision 0→1 acked by the kernel in 25–30 ms, propagated to `/openral/world_state_fast` (n_attached 0→1), no flapping. The Phase-2 "attached_objects stayed empty" gap is closed.

**Q2(a) — grasped-object voxels stopping robot links: did not recur.** No `panda_link*`-vs-payload-voxel stop in either run; the colliding side was always the attached payload itself. Depth transparency is doing its job.

**Q2(b) — support-contact stop: still fatal, reproduced 2/2.** `attached:sim:obj_main` vs `voxel_194478`/`voxel_194355`, min_distance ≈ −40 mm, ~5 s after attach, while MuJoCo ground truth shows the payload's only external contact is the counter top at 0.1–2.9 mm penetration (the `attached_payload_estop_snapshot` instrumentation made this diagnosable). Sim `attached_contact_tolerance_m` = 25 mm (launch sets it to octomap resolution; kernel default 1 mm) — misses by ~15 mm. Both runs are safety stops post-grasp/pre-lift; XR-1 grasped successfully both times, so these are not policy failures. 0/2 task completions.

**Settling-gated `action_applied` barrier: exonerated.** Kernel acks in ≤30 ms; the 8 s timeout messages are downstream of the E-stop (HAL stops applying after `estop_received`), not the barrier.

### Safety-WG agenda (calibration/semantics questions — no code changed to force a pass)
1. `attached_contact_tolerance_m`: kernel default 1 mm vs launch override to octomap resolution (25 mm sim; would be 50 mm at real 5 cm voxels). Resolution-scaled tolerance has a rationale (the false stop *is* voxel quantization error) but is a real permissiveness increase, esp. combined with `attached_contact_allow_new_shallow`.
2. The contact baseline is keyed by voxel **index** into a robot-centred, re-centring grid: as the base drives, indices decorrelate from world cells — the allowance can both fail to absorb a real support contact (observed) and in principle permit contact against different geometry (not observed, but structural).
3. Frame convention: the occupancy grid declares `frame_id: base_link` but its content is world/floor-referenced (counter surface at z≈0.89), while TF's `base_link` sits at z=0.70 in /odom — two `base_link` conventions coexist (echoes the #103 0.7 m pedestal issue). Kernel and grid currently agree with each other; any TF consumer of the grid is 0.7 m out. Under investigation; if the kernel side turns out wrong, the Q2(b) diagnosis flips.

### Conclusion
Raising the tolerance is not the fix; the **support-contact witness** (ADR-0092 follow-on) is: World State attests the payload's legitimate support contact at attach time and Safety exempts that bounded contact until separation. #102 must not close on this evidence. Full logs: `~/openral-runs/2026-08-12-attached-validation/` (runs 1–3, E-stop evidence snapshot `run3_estop_evidence.npz`).

https://claude.ai/code/session_01FLpBqPYFHRTVNQDcpGooaV

PR #123, opened 2026-08-15
feat(hal): ground-truth E-stop diagnostics and near-miss probes

AdrianLlopart:
## What changed

E-stop diagnostics that made every later verdict in this stack decidable.

- `sim.estop_ground_truth_snapshot` now fires on **every** kernel E-stop, not only when a payload is carried (the old `attached_payload_estop_snapshot`, renamed after confirming no consumer keyed on it). Carries MuJoCo robot↔world contacts, robot joint state, base-frame TF, the last three candidate-chunk digests, and the correlated `CollisionEvidence` (joined by `stop_seq` when it arrives late).
- Three bounded `mj_geomDistance` near-miss probes — robot↔world, payload↔world, payload↔robot — scoped to the manifest's kernel-checked links via `kernel_checked_body_ids`, with per-probe coverage attestation so an absent pair is only readable when the probe says it looked.
- Plane geoms get an exact lower bound instead of `-inf`, and the call budget is spent round-robin across the robot's geoms.
- `fix(sim,hal)`: the depth self-filter's **clearing** rays reach the octomap path again. The single-cast refactor (#111) derived the octomap cloud from the depth image, which discarded the clearing mask — making the robot's own silhouette a write-only region of the map (markable, never clearable).

## Why

The kernel stops on a **margin**, so a real stop usually leaves `ncon == 0`; a contact list cannot adjudicate it. Field proof: an arm 17.5 mm inside a RoboCasa freezer door produces zero MuJoCo contacts because the door's `contype` excludes it. Without the near-miss probe, three validation stops were unadjudicable and two earlier verdicts were wrong. The clearing-ray regression produced a frozen stale plane hugging `panda_link1` that tripped a false E-stop at tick 1.

## How tested

Real compiled `MjModel` fixtures throughout, no mocks: the `ncon == 0`-with-interpenetration pathology reproduced in a fixture; probe saturation (a wheeled base on a floor plane must still report arm-side pairs); payload↔world and payload↔robot probes; octomap persistence against a real `octomap::OcTree` proving the two-hit noise protection survives the clearing fix (one hit is not a safety voxel, two are, one clearing ray retires a confirmed voxel).

Unit suite green at each commit; `docs/methods/01-hal.md` and `14-duplication-watch.md` updated in range.


AdrianLlopart:
Superseded by the recovered chain (#128–#133). The original stack conflicted with master after #103 was squash-merged: squashing rewrote its eleven commits, so every downstream branch replayed originals master no longer recognised. The same content is now rebased onto master with the duplicates dropped (verified by diff, since `git cherry` marked all 57 commits as new), split across #128–#133, each independently verified. Superseded by #130.

PR #124, opened 2026-08-15
feat(safety): support-contact witness, payload clearing, and their partition

AdrianLlopart:
## What changed

The support-contact witness, payload occupancy clearing, and the partition that lets both coexist.

- **`SupportContactWitness`** (msg + schema + kernel): World State attests bounded support contact at attach; Safety exempts exactly that attested contact while it stays bounded and non-deepening, until separation. The attestation lives in the **object frame** (it follows the payload, so it cannot decorrelate as the base drives) and its `max_penetration_m` is **physical** — quantization is handled separately and geometrically, so the field can never be inflated into a license for real penetration. Hysteresis: separation kills the exemption, re-contact is a new violation, re-arm is keyed on `(object_id, support_id, stamp_ns)`.
- **Retires two weaker mechanisms**: the index-keyed non-deepening baseline (which decorrelated in exactly the case it existed for) and `attached_contact_allow_new_shallow`.
- **Payload occupancy clearing**: the octomap bridge zeroes occupied cells within one cell-circumradius of an attached payload's primitives, on every grid it publishes, honoring the message contract that already said *"the same object must be absent from world occupancy while attached."*
- **Distance-probe attestation basis** (`sim_geom_distance`): the producer measures support contact with `mj_geomDistance` instead of MuJoCo's contact list, which `contype`/`conaffinity` suppression silently empties for many object↔furniture pairs.
- **The partition**: clearing **withholds** cells inside the attested support patch — the withhold predicate is the kernel's exemption predicate at `slack = 0`, so `withheld ⊆ exempt` by construction and no withheld cell can ever be the cell that stops the robot. Those cells are real furniture the map should keep for Nav2 anyway.

## Why

Validation showed the payload-vs-world stop was real but its evidence uninterpretable, then that the payload's own silhouette tripped it, then — once clearing landed — that the witness starved on the very occupancy it needs as proof of contact. Each fix is a response to a measured failure, in that order.

**Closes HZ-0095-2**: `attached_contact_tolerance_m` returns to meaning 1 mm of physical slack; the 25 mm sim override is gone.

## How tested

Kernel gtests for every transition class (quantization-inflated contact passes, no attestation stops, deepening stops, separation-then-recontact stops, a wall cell 150 mm out stops while the support cell stays exempt, malformed attestation fails the whole message closed); `test_no_alloc` extended over the new latch, still zero allocations across 10,000 iterations. Bridge gtests on a real `octomap::OcTree` seeded through real `insertPointCloud`. A **53,872-probe** sweep mechanises `withheld ⊆ exempt` cell by cell. A kernel test proves payload-vs-world checking still stops against a real obstacle with the payload's own cells cleared.

⚠️ **Safety-WG**: hazard-log Entry 012 (HZ-0092-1/2/3) records the residuals — identity-blindness within the attested patch, the 12.5 mm discrimination floor at 25 mm voxels, the `/octomap_binary` silhouette residual, and the cross-package predicate-mirror maintenance obligation.


AdrianLlopart:
Superseded by the recovered chain (#128–#133). The original stack conflicted with master after #103 was squash-merged: squashing rewrote its eleven commits, so every downstream branch replayed originals master no longer recognised. The same content is now rebased onto master with the duplicates dropped (verified by diff, since `git cherry` marked all 57 commits as new), split across #128–#133, each independently verified. Superseded by #131.

PR #125, opened 2026-08-15
feat(safety): place-phase witness, attach-sweep window, and the barrier release fix

AdrianLlopart:
## What changed

The place phase (ADR-0097), the attach-transition sweep window, and the barrier fix that stopped a successful attestation from aborting its own goal.

- **Place-phase witness (ADR-0097)**: a typed `PlaceDeclaration` — named target, object, timeout, lifecycle — flows dispatch → HAL → World State on the same path attachment evidence already rides. While it is live, fresh bounded support contact within the declared target's subtree is attested and exempted with the *same* machinery, bounds, caps and hysteresis as the pick witness. **No declaration → the anti-scope stands unchanged**: a mid-carry wall-brush is still a violation, and `NewContactOutsideThePatchStillStopsMidCarry` passes verbatim. The kernel never sees the declaration — it stays producer-blind, which is what keeps its own rule evidence-based.
- **Lifecycle (HZ-0097-3)**: the declaration retracts on goal end, cancel, E-stop, and — a gap the tests found — in a `finally` around the execute callback, because rclpy aborts a goal on an escaping exception *without* unwinding per-goal state.
- **Attach-transition sweep window**: the initial clearing sweep runs at steady reach **plus one voxel**, and stays widened until the payload has moved one padding-distance from its attach anchor (measured in the OctoMap frame, since the residue is world-fixed and a driving base leaves it behind). Kills the stale pre-attach silhouette class.
- **Task-success signal**: `sim.task_success` on every edge of RoboCasa's own `_check_success`, plus a terminal verdict — the first time task completion is decidable from a validation run's artifacts. Both edges, because RoboCasa success is not latched; `None` is never reported as failure.
- **Barrier fix**: the release now keys on `(object_id, evidence_ref)` — the exact fact perception masks — instead of object-id addition. An attestation-only revision re-publishes the same object, so the old check saw no new geometry, never released the deferred action tick, and the goal died on the apply timeout 8 s later. Same fix closes a partial-detach sibling case.
- **Signal-teardown verdict** and an E-stop probe that no longer measures against non-collidable marker geoms (source of phantom penetration readings).

## Why

Placing into an enclosed, mapped container had no exemption path — validation stopped the payload 22–30 mm short of the target shelf every time. And once the place attestation worked, it *aborted the goal that produced it*: the single most ironic defect of the effort, reproduced 2/2 in the field.

## How tested

Producer 15/15 (declared/undeclared, wrong target still capped, non-load-bearing face attests nothing, unresolvable target refused, one-declaration-one-attestation, retraction/timeout/staleness). Dispatcher 7/7 on a real rclpy graph including the executor-escape case. Kernel: declared insertion rides the exemption while the same insertion undeclared still stops at −15.70 mm; the declared shelf is exempt while the cabinet's back panel is not. Barrier: a live-ROS test drives the production lifecycle node on a real `SimAttachedHAL` — RED shows `observed acknowledgements: []`, GREEN releases tick 180. Attach window: 9 gtests including the field's own 29-grid timeline.

⚠️ **Safety-WG**: hazard-log Entry 013 (HZ-0097-1/2/3) — inherited identity-blindness, wrong declared target, declaration lifecycle.


AdrianLlopart:
Superseded by the recovered chain (#128–#133). The original stack conflicted with master after #103 was squash-merged: squashing rewrote its eleven commits, so every downstream branch replayed originals master no longer recognised. The same content is now rebased onto master with the duplicates dropped (verified by diff, since `git cherry` marked all 57 commits as new), split across #128–#133, each independently verified. Superseded by #132.

PR #126, opened 2026-08-15
feat(safety): declaration-scoped approach allowance, clock fix, and quantization headroom

AdrianLlopart:
## What changed

The declaration-scoped approach allowance and the two calibrations that made it reach contact — **the tip that closes issue #102's acceptance criterion**.

- **Approach allowance (ADR-0097 amendment)**: the producer measures a `PlaceRegion` — an oriented box in the robot base frame, bounded at 1.5 m per side / 8 m³ — from the declared target's subtree, and ships it kernel-ward on the existing World State envelope. While the declaration is live, payload-vs-world cells **inside that region** get their margin reduced by `min(1.5 × resolution, 0.04)`. Reactive and predicted-horizon alike. Arm-vs-world, payload-vs-world outside the region, and the witness are untouched.
  It is a **margin change, not an exemption**: deepening past it still stops, the reported distance stays the pair's true distance, and `place_allowance_active` (log key + span attribute) discloses when a stop happened at a reduced margin. Fails closed toward *no allowance* — deliberately the opposite direction from a malformed attestation, because a bad region can only make the kernel more permissive.
- **Clock-domain fix**: declaration liveness is judged in the attachment stream's own clock. World State was evaluating it against wall time while the producer stamped in sim time, so every declaration read as expired and the region never reached the kernel. The unit test that should have caught it had injected a shared clock, testing away the only condition that fails in production.
- **Refusal taxonomy**: six distinct refusals no longer all report `reason=bounds`; the normal pre-grasp state (`no_object`) is INFO, transition-gated — 672–811 warnings per run became 3–4 lines.
- **Two one-voxel headrooms** (maintainer-directed, battery-evidenced): the allowance cap `min(1.5 × voxel, 4 cm)` — one voxel was structurally marginal because the map-vs-truth error at a placement pose is *itself* about one voxel — and the witness height envelope gains `+ resolution` within the lateral patch, for cells of adjacent co-planar structure. The bridge's mirrored withhold predicate moves in lockstep; the invariant sweep grew to 53,872 probes.

## Why

Issue #102's acceptance criterion. The chain was: the witness is earned by touching, but the kernel stopped the payload before it could touch (allowance), the region never arrived (clock), and the allowance was sized just under the discretization error (headroom).

## Result

At this tip the **baguette** scene — `PickPlaceCounterToCabinet`, the named case — completed pick→carry→place **twice independently** in a five-run battery, with no collision verdict in either run:

```
run2: sim.task_success first_success=True success=True step=575
      sim.task_success_final ever_succeeded=True steps=1265 transitions=1
run3: sim.task_success first_success=True success=True step=595
      sim.task_success_final ever_succeeded=True steps=1263 transitions=1
```

Kernel gate, Nav2 and SLAM active, XR-1 dispatched directly, no safety parameter touched. The remaining three runs carry `place_allowance_active=0` with `sweep_min == min_distance` — nothing exempted, so no witness or allowance question is posed: they are policy trajectory quality. The `fridge_drawer` control stop persists **byte-identically**, proving the headrooms did not absorb a real interpenetration.

## How tested

Kernel gtests 155/155 — the run-1 field penetration (−26.48 mm) now reaches contact declared and still stops undeclared; the +42.9 mm co-planar cell is exempt and +55 mm stops; a 50 mm grid caps at 40 mm, not 75 mm. Bridge 42/42 and the lockstep sweep. Dead-stream gate tests pin the two kernel gates a declaration expires through. Live-ROS delivery-seam test carries a sim-stamped declaration through the production node with the region intact.

⚠️ **Safety-WG**: ADR-0097 amendments + hazard Entry 013 HZ-0097-4 (pre-contact allowance exposure) and Entry 012's calibration record. Two honest caveats: the retired classes are demonstrated by counterfactual on recorded artifacts rather than a live firing, and the co-planar margin is 0.559 mm at sim resolution — follow-up tests for both are in flight.


AdrianLlopart:
Superseded by the recovered chain (#128–#133). The original stack conflicted with master after #103 was squash-merged: squashing rewrote its eleven commits, so every downstream branch replayed originals master no longer recognised. The same content is now rebased onto master with the duplicates dropped (verified by diff, since `git cherry` marked all 57 commits as new), split across #128–#133, each independently verified. Superseded by #133.

PR #127, opened 2026-08-15
test(safety): pin the witness height envelope to its four named terms

AdrianLlopart:
## What changed

Pins the thinnest safety-relevant margin in the collision stack — the support-contact witness's height envelope — against the exact field geometry that motivated its calibration.

`SupportContactWitness.TheWidenedEnvelopeIsExactlyItsFourNamedTerms`:

- **Derives** the envelope from its four named constants (`half·Σ|n| + attested_penetration + slack + resolution`) and pins derivation ≡ the test's literal, so a constant change *and* a formula change both fail loudly. The previous `kBatteryEnvelope = 0.0435` was a hardcoded literal that a constant moving underneath it would have slipped past.
- Pins that the derived value is the **floor** over five tilted support-face attitudes.
- Pins the round-8-r2 field cell (+42.893 mm, 85.7 mm lateral, inside the 135.65 mm lateral gate) as exempt, with its margin asserted numerically.
- Closes the bound: exempt at `derived − 1e-6` and exactly at `derived`, **not** exempt at `derived + 1e-6`.

Mutation-verified: a 25 µm change (`1.001 × resolution`) to the kernel formula fails the test.

## Corrects the record

The envelope at the field geometry (25 mm cells, 5 mm attested depth, 1 mm tolerance) is **43.500 mm exactly**, not the 43.452 mm the validation battery reported. That figure is arithmetically unreachable: it would require `Σ|n| = 0.996`, but that sum is ≥ 1 for every unit normal, with equality only when axis-aligned. **The true margin against the 42.893 mm cell is 0.607 mm, not 0.559 mm.**

## Bridge: no mirror needed, verified rather than assumed

A bridge mirror test was written, then its own justification was disproved and it was reverted. Mutation testing settles it: adding the kernel's slack term to `support_patch_withholds` fails `WithholdingIsTheKernelsExemptionPredicateAtZeroSlack` at 3 of its 7 normals, and an in-lockstep `2 × resolution` change fails 9 bridge tests. At the round-8 geometry the bridge's ceiling is 42.500 mm — 0.393 mm *below* the field cell — so the mirror clears what the kernel exempts. That is the safe direction, and `withheld ⊆ exempt` holds.

## How tested

Kernel 156 gtests (155 at base + 1); bridge colcon 42; unit suite 4639 passed / 109 skipped; clang-format clean. `refresh_methods_linenos --check` and `schema_export --check` are no-ops — no public symbol touched.

Note: `test_lifecycle_kernel` exits 139 at process teardown *after* all 21 of its tests report OK — the pre-existing vendored-OTLP shutdown issue on this host, unrelated to this change. Results counted from gtest output, not exit codes.

### DIFF
--- a/cpp/openral_safety_kernel/test/test_collision.cpp
+++ b/cpp/openral_safety_kernel/test/test_collision.cpp
@@ -1950,6 +1950,22 @@ constexpr double kBatteryEnvelope = 0.0435;    // 12.5 + 5 + 1 + 25 mm
 constexpr double kR2CoplanarHeight = 0.0429;   // round-8 r2's measured excess
 constexpr double kR2LateralOffset = 0.085;     // inside the run's 84-86 mm band
 
+// The kernel node's own physical FK/pose slack — `attached_contact_tolerance_m`,
+// declared and defaulted to 1 mm in `src/lifecycle_kernel.cpp`. It is the third
+// term of the envelope, so it is named here rather than spelled `0.001` at each
+// call site.
+constexpr double kBatteryContactTolerance = 0.001;
+
+// The SAME round-8 r2 cell to the micrometre, as the final battery recorded it
+// (`spark:~/openral-runs/2026-08-15-final-battery/`): +42.893 mm above the
+// attested support plane, 85.7 mm out. `kR2CoplanarHeight` / `kR2LateralOffset`
+// above carry the same cell rounded to the tenth of a millimetre, and the
+// rounding went UP, so the rounded fixture is the STRICTER of the two probes —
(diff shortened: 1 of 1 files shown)


AdrianLlopart:
Superseded by the recovered chain (#128–#133). The original stack conflicted with master after #103 was squash-merged: squashing rewrote its eleven commits, so every downstream branch replayed originals master no longer recognised. The same content is now rebased onto master with the duplicates dropped (verified by diff, since `git cherry` marked all 57 commits as new), split across #128–#133, each independently verified. Superseded by #133.

PR #162, opened 2026-08-24
fix(tests): make the opt-in robosuite lane actually run — and stop it prescribing a command it cannot run

AdrianLlopart:
## What changed

The `select-and-test` job's opt-in robosuite path failed for anyone whose diff
reached `python/hal/tests/`, with

```
ROSConfigError: openarm_robosuite needs a different robosuite than the one
installed, but this process has already imported it.
```

and separately told the reader to run `just sync …` on a container with no
`just`. Both are fixed; `_assert_no_live_dependency_swap` is untouched.

### 1. The guard was right; the lane assignment was wrong

`tools/test_selection.toml` picks a dependency lane **per file**.
`python/hal/tests/test_sim_attached_action_dim.py` held both LIBERO tests
(`robosuite==1.4`) and the OpenArm tabletop test (`robosuite>=1.5`, the
`robocasa` group) — mutually exclusive pins. In the `libero` lane the LIBERO
half imported robosuite 1.4 first, so `ensure_backend_deps("openarm_robosuite")`
found the wrong version and the guard correctly refused to swap the package
under live objects. The test **failed instead of running**, and ran correctly
nowhere else.

`tests/sim/test_openarm_scene_pnp.py` was in no lane at all — three tests that
had never executed on CI since they landed.

Fix: the OpenArm test moves to its own file on the `robocasa` lane, together
with `test_openarm_scene_pnp.py`. Batched, not isolated — the batched path is
the one that rejects a skip.

### 2. The selector dropped lane files that lived inside a directory target

`targets` mixes files with directories; a directory string never `fnmatch`es a
file glob. Any lane-owned file reachable only *inside* a selected directory
silently lost its lane: still "selected", but run only in the cheap partition
where it skips for want of the optional stack — selected, never executed, green.
Directory targets are now expanded to the lane files beneath them.

### 3. `just` was prescribed where `just` does not exist

`tools/verify_test_envs.py` shells out to `just sync` and died with a bare
`FileNotFoundError: 'just'` in the `robocasa-gr1` / `simpler-env` lanes; every
`_deps` hint names `just sync …` too. CI now installs `just` (ubuntu-24.04
universe), `verify_test_envs.py` preflights it with an actionable message
instead of a traceback, and `_deps.remediation()` appends the recipe's literal
expansion when `just` is genuinely off PATH.

## Why

Six-plus silent-degradation defects of the same family surfaced in this repo
this week: a plausible-looking wrong state produced quietly where a loud failure
was available. Two more here — a test that failed in the only lane that selected
it, and a whole file that skipped everywhere.

## How tested

- **The lane itself, for real**: `pytest
  python/hal/tests/test_sim_attached_openarm_action_dim.py
  tests/sim/test_openarm_scene_pnp.py tests/sim/test_so100_robosuite_lift.py`
  in a `--group robocasa` venv (robosuite 1.5.2, GTX 1060, EGL) — **18 passed,
  0 skipped**, one interpreter: exactly the batch and the strictness the lane
  applies. Run against `fcf7f01`, i.e. including #152's `robots/openarm/robot.yaml`
  changes.
- `tests/unit/test_select_tests.py` +4 cases; the directory-expansion one was
  confirmed to **fail** against the pre-fix `select_tests.py`.
- `tests/unit/test_deps_install_plans.py` +6 cases against a PATH narrowed to
  the real `uv` and nothing else, incl. one asserting the guard still refuses
  the swap while naming a runnable command; plus an anti-drift case that reads
  the real `Justfile`.
- `just lint`, `mypy --strict tools/ -p openral_sim -p openral_core`,
  full `tests/unit/`.

**Nothing is made to skip** — this strictly increases what executes: the OpenArm
HAL test now passes instead of failing, `test_openarm_scene_pnp.py`'s three tests
run for the first time, and a `robots/**` diff additionally reruns the `clip` /
`opencv` / `onnx-export` / `sidecar-wire` lanes it always implicitly selected but
never exercised.

## Not bundled

This PR touches `tools/test_selection.toml` + `tools/select_tests.py` +
`test-selective.yml`, all `full_run_globs`, so it forces a full run and every
lane — including lanes already failing on `master` for unrelated reasons
(`isaacsim`, `robotwin`, `gr00t` "produced no passing tests"; the `libero` /
`maniskill3` "still skipped selected tests" CUDA/Vulkan strictness). Those
predate this branch and are visible identically on run 32644213906 and PR #153's
run 32646951312.

### DIFF
--- a/.github/workflows/test-selective.yml
+++ b/.github/workflows/test-selective.yml
@@ -160,11 +160,24 @@ jobs:
       # test_cli_dataset_from_bag, …) fail with "Could not load libtorchcodec
       # … libavutil.so.* cannot open shared object file" whenever a broad diff
       # selects them. Install FFmpeg (provides libavutil.so.58 → core6).
-      - name: Install native runtime libs (torchcodec + MuJoCo OSMesa)
+      #
+      # `just` is installed for TWO reasons, both load-bearing:
+      #   1. tools/verify_test_envs.py (the robocasa-gr1 / simpler-env lanes)
+      #      shells out to `just sync --group <g>`. Without the binary those
+      #      lanes died with a bare `FileNotFoundError: 'just'` traceback
+      #      before running a single test.
+      #   2. Every optional-dependency failure in openral_sim._deps prints
+      #      "Install it first, then re-run: just sync --all-packages --group
+      #      <g> --inexact". A remediation you cannot run in the environment
+      #      that printed it is worse than no remediation — it sends the next
+      #      reader down a dead end. CLAUDE.md §1.2 (truth over plausibility).
+      # ubuntu-24.04 ships just 1.21 in universe, so no extra action/tap.
+      - name: Install native runtime libs (torchcodec + MuJoCo OSMesa + just)
         if: steps.select.outputs.any == 'true'
         run: |
           sudo apt-get update -qq
-          sudo apt-get install -y --no-install-recommends ffmpeg libosmesa6
+          sudo apt-get install -y --no-install-recommends ffmpeg libosmesa6 just
+          just --version
 
       - name: Install Python dependencies
         if: steps.select.outputs.any == 'true'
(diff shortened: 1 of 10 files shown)


AdrianLlopart:
-

AdrianLlopart:
-

Issue #176, opened 2026-08-26
A declared place ends in an E-stop: payload↔world contact has no outcome between "allowed" and "latched"


AdrianLlopart:
## Summary

A carried payload touching the world during a **declared place** is a latched
E-stop, at any depth, exactly like an arm link crashing into a cabinet. Those
two are not the same event and the stack already knows the difference — it just
does not act on it.

Measured on the 2026-08-26 oriented-grid round (`baguette`): the kernel stopped
at **−38.22 mm** with `place_allowance_active=true`, **0.72 mm** past its
allowance ceiling. The ground-truth probe puts the physical contact at
**−1.4 mm**. A 1.4 mm touch of the receptacle ended the run, and needed
`/openral/estop_reset` to recover.

## What exists today

There is no grasp/carry/place mode. There are three independent facts:

| Fact | Carrier | Producer |
| --- | --- | --- |
| carrying | `WorldStateStamped.attached_objects` + `attachment_revision` | HAL attachment evidence |
| resting on a support | `SupportContactWitness` (object frame, dies on separation) | World State |
| placing | `PlaceDeclaration` (`target_id`, `timeout_s` backstop) | rSkill runner, at goal start (ADR-0097) |

The separation this issue needs already exists in the kernel: robot links go
through `check_voxel_collision`, payloads through
`check_attached_voxel_collision`, and `place_approach_allowance` is applied
**only** to the payload path. **Nothing here proposes touching the robot-link
case.**

What is missing is on the outcome side. Every geometric hit — link or payload,
declared place or not — funnels into the same `report()` in
`lifecycle_kernel.cpp`:

```
fault_latch_ = true;  publish /openral/estop;  drop the chunk;  severity = "violation"
```

There is no warning severity, no refuse-without-latching, no graded response of
any kind. Recovery is `/openral/estop_reset` plus a cooldown.

## Why the current allowance does not cover a place

Sim runs `world_voxel_margin_m = 0.0` and `attached_collision_margin_m = 0.0`,
so the gate is `d <= margin - allowance`:

* **payload↔world, no declaration** — trips at `d <= 0`. *Any* contact latches.
* **payload↔world, inside its declared region** — allowance is
  `min(1.5 x voxel, 40 mm)` = **37.5 mm** at 25 mm cells, so it trips at
  `d <= -37.5 mm`.

The allowance is a **margin reduction**, calibrated (ADR-0097's 2026-08-14
[...]

Issue #174, opened 2026-08-25
The depth path voxelises non-collidable geometry: #149 filtered visual geoms out of the adjudicator, not out of the map


AdrianLlopart:
Split out of #171 so it survives as work rather than as prose in a study.

## The asymmetry

#149 established that a geom with neither `contype` nor `conaffinity` cannot collide with anything, and taught the **ground-truth probe** to exclude those geoms — a `0.000 m` pair against a visual mesh can no longer be promoted to a `real_contact`, and the coverage block reports `noncollidable_*_geoms_excluded` so the omission stays visible.

That fixed the **adjudicator**. It did not touch the **perception path**, and nothing else does either:

* `mj_multiRay` in the depth synth strikes whatever is *visible*; it has no collidability filter (correctly — a depth sensor has no such notion).
* The synthesised cloud back-projects those returns, octomap integrates them, and `openral_octomap_bridge` rasterises them into `/openral/world_voxels`.
* The C++ safety kernel treats every occupied cell as an obstacle.

So a visual-only geom becomes an obstacle the kernel will E-stop on, while the probe that adjudicates that same stop is required to ignore it. The two halves of the stack disagree about what counts as world geometry.

## Scale

Measured in [`docs/reference/world-map-fidelity.md`](docs/reference/world-map-fidelity.md):

* **713 of 1 714** world geoms in the RoboCasa fridge scene are non-collidable decoration — 42 %.
* In a uniform sample of 1 200 of the 9 667 occupied cells of one live map (`robocasa_baguette` layout 7), **225 cells (18.8 %)** are backed *only* by a non-collidable geom — real to the camera, absent from physics.

This is one of the two reasons the live map is **more** conservative than the geometrically ideal grid, not less: over 59 fridge layouts the live map reproduces all 55 census stops, adds 2, and is on median 13.04 mm more negative. The ideal grid is built from collidable surfaces only, so it cannot contain these cells at all.

## Why it is sim-specific, and why that still matters

**On real hardware this term does not exist.** A visible object in reality *is* solid; there is no such thing as a geom the camera sees and the world cannot touch. The artefact is entirely a consequence of MJCF carrying decorative geometry alongside collision geometry.

That makes it a **sim-fidelity** problem rather than a safety problem, and it cuts in the conservative direction, so nothing is unsafe today. But it distorts every sim-derived number the collision programme rests on:

* it inflates live stop counts relative to what the same scene would produce on hardware;
[...]

Issue #173, opened 2026-08-25
The octree→grid cube-overlap rule holds 48% of live start-state stops


AdrianLlopart:
Split out of #171 so it survives as work rather than as prose in a study.

## What was measured

[`docs/reference/world-map-fidelity.md`](docs/reference/world-map-fidelity.md) apportions 23 live start-state stops by re-scoring each against the *same* published `/openral/world_voxels` map with exact link geometry (the quantity #166's staged 26-DOP → exact-hull pipeline reaches). A stop that exact geometry clears is link-side; one it keeps is world-side by construction.

Result: **26 % link-side, 74 % world-side.** Splitting the world-side residue by what backs the cell that stops exact geometry — the repo's own 243-ray `voxel_backing_record` probe and octree membership agree on **17 of 17**:

| world-side term | n | of all stops |
| --- | ---: | ---: |
| quantisation — real collidable surface inside the cell, cell present in the octree | 5 | 22 % |
| **cube-overlap dilation** — cell holds no world geometry, centre in **no** octree leaf, real surface ≈ one cell away | **11** | **48 %** |
| sensor-side phantom — nothing collidable within 0.29 m | 1 | 4 % |

## The mechanism

`rasterize_octree_to_grid` (`packages/openral_octomap_bridge/src/octree_to_grid.cpp`) marks a base-frame cell when its cube **shares volume** with an occupied octree leaf's cube. The two lattices share a resolution but not a phase — and on a mobile base not a yaw either — so a leaf holding a real surface also marks cells up to one full cell away from it.

A cell whose centre lies inside no occupied leaf can only have come from that rule, and **12 of 12** non-`solid_world` tripping cells are exactly that, each 16.17–30.52 mm from the nearest occupied leaf.

This identifies the mechanism behind #160's "empty cell with solid cabinet exactly one cell away". The phantoms are **displaced, not stale or spurious**:

* not stale — every capture is a start state with zero actions applied, and the depth self-filter makes robot bodies transparent, so the arm cannot have written them;
* not spurious — octomap needs two hits to lift a cell over `occupancy_thres` (0.8) and one clearing ray to drop it back, so noise cannot persist anywhere a ray passes;
* displaced by exactly one cell, deterministically.

## Why this is not a bug report

**The rule is deliberate and is the safe direction.** Its own comment records that the previous centre-sampling rule quantised every surface to whichever lattice the cell centre landed on, displacing surfaces by a whole voxel *toward* the robot. Overlap fixed that.

[...]

PR #179, opened 2026-08-27
feat(safety)!: refuse a declared place's own contact instead of latching an E-stop

AdrianLlopart:
Closes the **outcome** half of #176. Stacked on #178 (`feat/oriented-world-grid`) —
review that one first; this diff is only the 671 lines on top.

## The problem

A carried payload touching the world during a **declared** place was a latched
E-stop, at any depth, exactly like an arm link crashing into a cabinet. From the
`oriented-grid-2` round: the baguette place read **−38.22 mm** against a 37.5 mm
approach allowance — 0.72 mm over — while the ground-truth probe put the physical
contact at **−1.4 mm**. A 1.4 mm touch of the receptacle ended the run and
required `/openral/estop_reset`.

That reading is inflated for reasons that are all map discretisation and none of
them force: `insertPointCloud` marks the cell *containing* the ray endpoint, so
the kernel's surface starts up to one resolution early, and the payload is a
fitted convex primitive that bulges past the real object.

**That error is a bound, not a bias — it cannot be subtracted**, because the same
−38 mm is also what a genuinely deep collision looks like. So this does not relax
the threshold. It adds an outcome *between* "allowed" and "latched", for one
narrowly-scoped class.

## What changed

`CollisionHit::advisory`. When set, the node drops the candidate chunk **without**
latching a fault and **without** asserting `/openral/estop`, logs
`safety.collision_advisory` at WARN, and marks the span `severity=advisory`.
`hit` stays true: it is a refusal, not a pass.

Five bounds, **any one of which failing gives today's latched stop**:

| bound | what it means |
| --- | --- |
| payload only | set inside `check_attached_voxel_collision` and nowhere else, so arm-vs-world is untouched at any depth |
| declared region only | requires `place_approach_allowance > 0`, returned only for a cell inside the producer-measured region for the declared object |
| bounded in depth | `min(0.2 × voxel, 5 mm)` past the allowance |
| bounded in repetition | `place_advisory_max_consecutive` (3) caps an unbroken run and the next refusal latches; any accepted chunk resets it, so the cap counts shoving rather than progress. **0 disables the band entirely** |
| severity outranks depth | in `fold_pair` |

## Two findings that shaped it

**The band had to be sized from measurement, not from a voxel multiple.** A probe
over the real routines shows per-cell penetration **saturates** near the allowance
itself — deepest reachable is −37.50 mm at 25 mm cells and −50.00 mm at 50 mm. The
first draft used one voxel (30 mm), which would have covered the entire reachable
range and left the depth limit bounding nothing. 0.2 voxels is ~7× the measured
0.72 mm overshoot and leaves real depth latching on both lattices.

**Ranking tripping pairs by depth alone becomes wrong the moment `advisory`
exists.** The gate threshold is no longer the same for every pair in a sweep, so
the deepest is not the most serious: a −5 mm hard trip on an undeclared payload
and a −43 mm advisory reading inside the declared receptacle can occur in one
sweep, and reporting the deeper one would hand the caller an advisory hit and
**silently lose the latch the first pair earned**.

## How tested

196 kernel tests. Eight new `PlaceAdvisoryBand.*` pin each bound — including that
a robot link is never advisory *however shallow*, and that the band stays a small
fraction of the allowance. Two new lifecycle tests drive it end to end:
`AnArrivedPlaceRefusesTheChunkWithoutLatching` asserts no latch and no
`/openral/estop`; `AnUnbrokenAdvisoryRunLatchesAtItsCap` asserts exactly three
advisories then an ordinary stop.

Both new lifecycle tests deactivate + clean up so `shutdown_tracing()` drains
their spans. Without that the process segfaulted at teardown inside the OTLP
exporter's background thread racing an absent collector — a latent condition the
extra spans were enough to lose.

## What the five-round battery says about this change — read this before approving

A five-round, four-scene battery ran on this exact tree
(`docs/reference/collision-validation-evidence.md`). **The band never fired
once**, in any of the 20 scene runs. Of the 15 stops:

* **9 are a robot link**, not the payload — outside this PR's scope entirely, and
  the real target of #176's remaining "quantisation at source" piece (#173, #174);
* **4 of the 6 payload stops had `place_declaration_seen: false`** — carried, never
  placed, so no version of this change could reach them;
* **2 are the case #176 actually describes** — declared, region armed, contact
  outside the armed volume.

So this ships **correct and measured-inert**. It revises my own earlier n=1 read,
which had over-weighted the region boundary as the binding constraint; it is a
minority case, 2 of 15. The honest summary is that this closes a real gap that the
current four scenes rarely exercise, and the majority failure class is elsewhere.
That is an argument about priority, not about correctness — but a reviewer should
know the band has no positive live evidence behind it, only test evidence.

## Notes for review

**This needs the CLAUDE.md §3 safety path and does not yet have it.** It weakens
an outcome the kernel previously always latched. Outstanding, and not
self-approvable:

- [ ] safety-WG reviewer
- [ ] hazard-log entry in the private `OpenRAL/management` repo
- [ ] sign-off on the five bounds above as the correct scope

`place_advisory_max_consecutive: 0` restores the previous behaviour **exactly**,
which is the intended rollback if the WG wants it off by default.

BREAKING CHANGE is declared on the commit: `CollisionHit` gains `advisory`, and a
payload contact inside a live place declaration within `place_advisory_depth` of
the allowance no longer latches or asserts E-stop.


AdrianLlopart:
## This PR's green CI is vacuous — here is the real verification

Worth stating plainly before any §3 sign-off: **the three green checks on this PR ran none of its tests.**

This PR's entire diff is `cpp/openral_safety_kernel/**` plus docs. `tools/test_selection.toml` excludes `cpp/**` from the Python selector, so `select-and-test` correctly reported:

```
##[notice]No code paths affected by this diff — no tests selected.
```

The toml says that exclusion is fine because `cpp/` is "covered by their own workflows: `test-ros2` (colcon) and the safety-kernel ctest". **That workflow does not exist** — no workflow in the repo runs `colcon test` or `ctest`. #182 adds it (dispatch-only, since the repo is out of Actions credits).

So this PR's CI signal is DCO + ruff/mypy/mkdocs. Nothing has exercised the advisory band in CI, and nothing will until #182 is PR-triggered and added to `required_status_checks`.

## What was actually run, locally, on this PR's rebased tree

`fec669e`, ROS 2 Jazzy:

```
colcon build --merge-install --symlink-install --base-paths packages cpp \
  --packages-up-to openral_safety_kernel openral_octomap_bridge
  Summary: 4 packages finished [2min 21s]

colcon test --merge-install \
  --packages-select openral_safety_kernel openral_octomap_bridge
colcon test-result --verbose
  Summary: 253 tests, 0 errors, 0 failures, 0 skipped
```

Read from the gtest XML rather than the summary line, so the suites are known present rather than silently absent:

| suite | tests |
| --- | ---: |
| `PlaceAdvisoryBand` | 8 |
| `OrientedGrid` | 2 |

That is the first execution of this branch's content since the rebase onto `423044a` (which now carries #180).

## What this does and does not establish

**Does:** the eight bounds tests and the two lifecycle tests pass on the exact tree proposed for merge, on a base that includes #180.

**Does not:** make this reproducible by anyone but me on demand. It is one run on one host, recorded in a comment. That is weaker than a CI artifact and should be treated as such — it is evidence for the reviewer, not a substitute for the missing lane.

It also does not change the [five-round finding](https://github.com/OpenRAL/openral/pull/179): the band never fired in 20 scene runs. Passing tests and inert-in-practice are both still true.

PR #204, opened 2026-09-04
fix(safety): convert the support-contact witness probe off mj_geomDistance

### LINKED ISSUE #190: fix(safety): convert the support-contact witness probe off mj_geomDistance
## Problem

#170 established that `mj_geomDistance` returns **confidently wrong values** for RoboCasa-fixture-vs-`panda_mobile`-mesh pairs in two silent modes (a `+0.148 mm` truth read as `0.000 mm`; four `0.000 m` readings re-measured at **+14.8…+107.9 mm** by the certified instrument), replaced it on the **evidence** path, and withdrew every verdict from any prior probe (validation-evidence ledger caveat 8).

**The support-contact witness path was never converted.** `_probe_support_hits` (`python/hal/src/openral_hal/_sim_attachment_evidence.py:616`, live call at `:649`) still measures payload↔support penetration with `mujoco.mj_geomDistance` — and unlike the evidence path, this one is a **safety input**: the `SupportContactWitness` it produces earns a kernel exemption (ADR-0092 D6, #131). #170's measured failure direction is *toward closer*, which on this path means **attesting a contact that is not there**.

Honest caveats already on record (survey §11 item 4): the probe runs at a 1 mm window rather than the 0.1 m windows #170 characterised, and #170 measured 1101/1102 pairs agreeing at the shipped windows — so the exposure is likely small. But #170 is explicit that *"no `distmax`, no 'only trust it below N mm' rule and no `ncon` cross-check separates the good answers from the bad"*, which is precisely the reasoning a narrow window would rely on. A known-defective instrument on a path that grants exemptions, unflagged in the evidence ledger, is the finding.

## Proposed solution

Convert the probe to `openral_hal.convex_distance.convex_geom_distance` — the same certified, self-checking instrument (witness-clearance + separating-axis certificates, attesting `certified_pairs`) that #170 installed on the evidence path. Keep `_SUPPORT_MAX_PENETRATION_M` (0.01, `:51`, enforced `:896`) and `_SUPPORT_PROBE_GAP_M` (0.001, `:45`) and their hand-maintained kernel mirrors (`docs/methods/14-duplication-watch.md`) unchanged — this is an instrument swap, not a semantics change.

## Verification

- Re-run #131's **53,872-probe invariant sweep** (`withheld ⊆ exempt` partition) with the certified instrument; every probe attests `certified_pairs`.
- Regression pinning at least one of #170's four re-measured `0.000 m` pairs on **this** path.
- One live round on `robocasa_baguette`: witness issuance timing/behaviour unchanged where it legitimately fires — the task still completes through the attach-sweep window (#132/#133 machinery untouched).
- Evidence ledger updated: the witness path's instrument lineage recorded so caveat 8's scope is explicit.

## References

- PRs: #170 (the two silent failure modes + certified replacement), #131 (witness + invariant sweep), #128, #132
- Files: `python/hal/src/openral_hal/_sim_attachment_evidence.py:616,:649,:45,:51,:896`; `docs/methods/14-duplication-watch.md`
- Survey §11 item 4; `docs/reference/collision-validation-evidence.md` caveat 8; snapshot: OpenRAL/management#18

AdrianLlopart:
Closes #190.

## Plan
- Extend the certified instrument's SAT branch with a witness pair for overlapping cores (the witness path needs a contact point on the penetrating branch; the evidence path did not).
- Convert `_probe_support_hits` to `convex_geom_distance` at the same 1 mm window; only certified measurements produce a hit.
- Keep `_SUPPORT_PROBE_GAP_M`, `_SUPPORT_MAX_PENETRATION_M` and their kernel mirrors unchanged.
- Record the lineage in the witness's `evidence_ref`, the ledger (new section + caveat 8 scope), the duplication watch and the method docs.
- Pin two of #170's pairs through the probe itself; re-run the #131 sweep; one live baguette round.

## What changed
- `convex_distance.py`: `_sat_penetration_depth` returns its minimum-translation axis (oriented so `a` is on its positive side) plus a witness pair — the centroid of `a`'s buried face, lifted onto `b`'s supporting plane — so `(witness_a − witness_b) / distance_m` is the `b`→`a` direction on **both** branches. Ball radii are applied along the axis. The evidence path's records gain `witness_*_xyz` on SAT pairs for free.
- `_sim_attachment_evidence.py`: `_probe_support_hits` measures with `convex_geom_distance(..., distmax_m=_SUPPORT_PROBE_GAP_M)`; beyond-window and uncertified pairs (a plane, an over-budget hull) attest nothing, so the failure direction is a *missing* exemption, never a false one. `evidence_ref` is now `openral_hal.convex_distance.convex_geom_distance:<support>`; `mujoco_geom_distance:<support>` identifies a witness from the defective instrument. Caps and kernel mirrors untouched.
- Docs: ledger section "2026-09-03 — the witness path converted off `mj_geomDistance` (#190)" and caveat 8 now names the witness path in its scope; `14-duplication-watch.md` entry renamed and forbids going back on either path; `01-hal.md` witness semantics; the stale call-budget cost comment replaced with measured numbers. A prior `fix(docs)` commit refreshes 22 pre-existing stale METHODS markers (§1.15).

## Why
#170 showed `mj_geomDistance` returning confidently wrong values *toward closer* on RoboCasa-fixture-vs-`panda_mobile`-mesh pairs and replaced it on the evidence path. The witness probe was never converted, and its output earns a kernel exemption (ADR-0092 D6). #170: no window separates the good answers from the bad — and the new pin below shows the false zero **inside the 1 mm witness window**.

## How tested
- `tests/unit/test_convex_distance.py`: new analytic test for the penetrating witness pair (both witnesses on their own surfaces, direction contract, JSON record). Unit: 81 passed across the witness/instrument/attachment files; evidence-path record tests 87 passed.
- `tests/sim/safety/test_support_probe_instrument_robocasa.py` (new, real kitchens, no mocks):
  - 08-23 `baguette` stop reconstructed (base residual 0.5 µm vs its own `base_frame_tf`; coincident `robot0_g12_vis` reproduces the round's `0.107931 m` to 2e-8 m). **The recorded `0.000` does not reproduce at the 1 mm window on this host** — the call saturates to `distmax` there — so that pin holds "no attestation" under both instruments. Stated as such, not oversold.
  - Layout-9 knife-edge pair (`robot0_link7_collision` ↔ `fridge_right_group_freezer_door_main`, certified +0.1485 mm): `mj_geomDistance` returns `0.000` **at the probe's own 1 mm window** with a 126 mm witness segment 0.53 / 0.43 m outside both geoms, and the pre-#190 loop replayed on it produces a phantom support hit (penetration 0.0, contact point in empty space, a face normal read off the door at a point not on the door). The converted probe attests nothing: the certified witness lands on the door's edge and the face-normal cross-check fails closed.
  - Both sim files: 9 passed. With the overlay sourced, the ROS-message tests that skip in a bare venv also pass (64 across the place-phase witness, attached-object and stamped-objects files).
- #131 sweep `PayloadClearing.WithholdingIsTheKernelsExemptionPredicateAtZeroSlack` (7×4×4×37×13 = 53 872 cells): re-run green with the whole bridge and kernel gtests (273 tests, 0 failures). It is a predicate-vs-predicate mirror on synthetic cells and never calls a distance instrument, so it is unaffected by construction — re-run rather than cited as coverage.
- Cost, measured on the shipped `robocasa_baguette` kitchen: 353 pairs pass the 1 mm prefilter, 234 rejected by the certified window unsolved, 119 solved, **0.21 s per attestation** (median of five, unloaded host). It runs at attach and then on **every tick of a live place declaration until one attests** — the hysteresis short-circuits only after a success. The place phase measures against the declared target's own bodies rather than the whole kitchen, which is what keeps that affordable. Attests `counter_1_left_group_main` at 0.065 mm with a 0.117 m patch and an exactly vertical world-frame normal.
- Live round `2026-09-03-190-witness-2` (harness, this host, `OPENRAL_ALLOW_REMOTE_CODE=1`, `--force-shared-gpu` because the user's own unit-test session held ~200 MiB): `safety.support_witness_armed object=sim:obj_main support=sim:counter_1_left_group_main` at attach, `support_witness_separated` 4.4 s later on lift — the #132/#133 window machinery untouched. The place did not complete: one advisory-band refusal at −4.49 mm against the declared cabinet geometry (`consecutive=1/3`, no latch, no E-stop snapshot), then the policy ran out its budget. A master-baseline run of the same scene on this host is recorded below for comparison.

- Master baseline `2026-09-04-190-master-baseline-1` at `ed850ac`, same scene/seed/host: **same outcome class**. Witness armed on the same counter at attach and separated on lift; one advisory-band refusal against the declared cabinet at −1.79 mm (vs −4.49 mm on this branch), no latch, policy ran out its budget (651 vs 469 steps; master re-grasped once). The place-phase failure is the policy's, on both instruments; n=1 per arm.

- `ruff`, `ruff format --check`, `mypy --strict` on both modules, `refresh_methods_linenos.py --check` clean.

## Code review, and the defect it found

A review of this PR found three issues; all three were reproduced before being fixed (commit `bdbb9d2`).

- **A wrong support plane, and this one matters.** The certified witness on the *overlapping* branch lies in the support's supporting **plane**, which is the right depth but need not fall within the support's face. On a counter built from coplanar strips, a payload flush on one strip yields — for a *neighbouring* strip — a support point 0.25 m outside it, and the box's analytic face normal there comes back **lateral**. `_dominant_support` groups by support root and **averages** the normals, so one such hit tilted an attested support plane **45 degrees off vertical**. That plane is the basis of a kernel exemption. Reproduced on a two-strip counter (they cancelled on a symmetric layout and did not on an asymmetric one).
- **`witness_b` was documented as "nearest point on geom `b`"**, which is false on that branch, and the unit test asserting it passed only because its two boxes were aligned.
- **The call-budget comment claimed the probe never runs per tick.** `_place_phase_transition` → `_place_witness` re-probes on **every** tick of a live place declaration until one attests, since the hysteresis short-circuits only after a success.

**Fix.** `convex_geom_distance` now reports its own contact `direction` — the separating direction, or the minimum-translation axis when overlapping. This is exact at a flush contact, which is precisely where differencing the two witness points yields nothing because they coincide. The probe cross-checks *every* hit's surface normal against it and drops the ones that disagree, so the agreement check now runs at the resting contact this module exists for instead of being skipped there; `_SEGMENT_DIRECTION_MIN_M` is gone with it. Both false claims corrected in the code, the ledger and the method docs.

**Live round on the fixed commit** (`2026-09-04-190-crosscheck-2`, at `bdbb9d2`): the witness arms on the same support (`support_witness_armed object=sim:obj_main support=sim:counter_1_left_group_main`), which is what a wrong cross-check would have removed. That round ran further than either earlier one (713 steps vs 469 / 651) and ended in a **real E-stop of the carried payload against world occupancy** — `attached:sim:obj_main` vs `voxel_232330` at −14.6 mm, `estop-collision-within-quantization`, no exemption active at the trip. It logged no `support_witness_separated`, unlike the two earlier rounds, though the payload had left the counter. That is a rollout difference, not an effect of this fix: the producer emits a **bit-identical** witness on this scene before and after it (same support, 0.065 mm, 0.117 m patch, same normal), because the tilt only bites where a neighbouring support geom sits laterally offset under the payload, which this scene's reset geometry does not present. Three rounds of one scene settle nothing about rates either way.

**Verified after the fix:** the 45° tilt is gone (new regression test `test_a_neighbouring_strip_cannot_tilt_the_attested_support_plane`); the real baguette kitchen still attests its counter with an **exactly vertical** world-frame normal; the honest `witness_b` contract is now pinned by its own test rather than implied; full unit suite 5297 passed; sim safety 13 passed including both RoboCasa pins.

## Not done here
- Uncertified supports (a plane floor) now attest nothing where the old probe attested from `mj_geomDistance`. Pinned by `test_a_plane_support_attests_nothing_because_it_cannot_be_certified` so the behaviour is stated, not silent. RoboCasa has no plane supports under a payload; a plane-aware branch in the instrument is a follow-up if a scene ever needs it.
- Hull decode per call is a large share of the 0.21 s (`ponytail:` note at the budget constant); cache per (model, mesh) if attach latency ever matters.

https://claude.ai/code/session_01HSutpDDKU5QufdAPtubnaP


AdrianLlopart:
### Requesting Safety-WG eyes, even though CODEOWNERS does not mandate them here

`.github/CODEOWNERS` scopes `@openral/safety` to `packages/openral_safety/**` and `cpp/openral_safety_kernel/**`. This PR touches neither — it changes `python/hal/src/openral_hal/`. But what it changes is the **producer of the `SupportContactWitness` the kernel grants an exemption on** (ADR-0092 D6, #131), so the blast radius is a safety input even though the path is not a safety package. Flagging rather than assuming.

Three things a safety reviewer should look at specifically:

1. **The failure direction is now "no exemption".** A hit is produced only from a *certified* measurement, so an unmeasurable support (a plane, an over-budget hull) attests nothing and the kernel stops on the contact. That is strictly more conservative than the previous behaviour, which attested from `mj_geomDistance` regardless. The one live behaviour change worth naming: a payload resting on a **plane** floor previously earned an exemption and now does not. No RoboCasa scene rests a payload on a plane; `test_a_plane_support_attests_nothing_because_it_cannot_be_certified` pins it.

2. **A wrong attested plane was found and fixed during review** (`bdbb9d2`). On a counter built from coplanar strips, a payload flush on one strip read a *lateral* face normal off a neighbouring strip — at a witness point 0.25 m outside that strip — and because `_dominant_support` averages a group's normals, the attested plane tilted **45° off vertical**. The fix is that `convex_geom_distance` now reports its own contact direction, exact even at a flush contact where the two witness points coincide, and every hit is cross-checked against it. Regression test: `test_a_neighbouring_strip_cannot_tilt_the_attested_support_plane`.

3. **Hazard log.** I have not written a hazard-log entry, because that is a Safety-WG artifact and I did not want to author one unilaterally. The question for the WG: does Entry 012 (HZ-0092-1/2/3) want a line recording that the witness instrument changed, and that **every witness issued before this PR rests on the defective one**? Standing caveat 8 in `docs/reference/collision-validation-evidence.md` now says so, and a witness whose `evidence_ref` begins `mujoco_geom_distance:` is one of them.

Kernel caps (`support_witness_max_penetration_m`, `support_witness_max_patch_radius_m`) and their hand-maintained producer mirrors are untouched, as is the 1 mm probe window.

PR #235, opened 2026-09-07
feat(safety): tight geometry for the carry-phase links, and the ceiling that justifies it

AdrianLlopart:
Three commits: the measurement that reframed the programme, the fix it pointed at, and the record.

## 1. The ceiling — the number nobody had taken

After a month of collision work, completion went 25 % (2026-08-26) → 5–10 % (2026-09-06), and **no round, census or survey had ever run the policy with the world-voxel gate off.** The harness *refuses* the flag (`_SAFETY_KNOB_PATTERNS`) — correctly for a validation round, and exactly why the number was missing.

`spark`, 88 valid runs, 4 scenes × 2 arms, 10–12 per cell, same commit and host, **both arms simultaneous** so contention loads onto each equally:

| | valid runs | completed | rate |
| --- | ---: | ---: | ---: |
| gate **OFF** | 45 | **14** | **31.1 %** |
| gate **ON** (shipped) | 43 | **1** | **2.3 %** |

**Fisher p = 3.5 × 10⁻⁴**, power 0.97. Leave-one-scene-out holds it.

| scene | OFF | ON | p |
| --- | ---: | ---: | ---: |
| `utensil` | 7/12 (58 %) | 0/10 | 0.005 |
| `fridge` | 5/11 (45 %) | 0/10 | 0.035 |
| `sink_cup` | 2/11 (18 %) | 1/11 (9 %) | 1.0 |
| `baguette` | 0/11 (0 %) | 0/12 | 1.0 |

**This does not license turning the gate off** — 6 of 91 stops in the #204 battery were real contact. It is a ceiling: the geometry levers are competing for **up to 29 points**, not for noise.

**`baguette` should leave the collision scorecard**: 0 % with the gate off means it is policy-bound and cannot report on collision work either way — despite four of the five completions in the ledger's history being baguette runs, which is what made it look like the bellwether.

## 2. The decomposition that reordered the levers

For every stop the battery records the kernel's reported depth *and* the certified mesh gap. Subtract the 21.65 mm cell half-diagonal:

| stop class | n | median excess | **beyond the voxel term** |
| --- | ---: | ---: | ---: |
| payload | 62 | 20.1 mm | **−1.5 mm** |
| link | 29 | 54.8 mm | **+33.1 mm** |

The payload primitives are **already tight** (`extract_body_primitives` lowers each geom separately, not one box) — so the plan's original "payload as a tight hull" lever had no headroom and is struck. The **link envelopes** were not tight.

## 3. The fix: tight geometry for `panda_link3`, `link4`, `link6`

`panda_link6` alone dominates **18 of the 29 link-class stops**, including seven on one fridge cell (`voxel_352030`) at 37–63 mm of certified clearance — and it shipped no `tight_geometry` at all.

| link | box excess → DOP | recovered |
| --- | --- | ---: |
| `link3` | 75.57 → 23.84 mm | 51.7 mm |
| `link4` | 76.12 → 23.15 mm | 53.0 mm |
| `link6` | 52.70 → 21.53 mm | **31.2 mm** |

`link6`'s 31.2 mm is almost exactly the measured 33.1 mm of link-class excess — the corroboration that the OBB slop *is* that excess.

**This reverses a recorded decision.** `collision-hull-narrow-phase.md` §5.2 excluded these three because #159 found they hold "zero of the 72 census stops". That census measured **start states** and was right about them; the #204 battery measured the **carry phase**, which is 71 % of stops, and there `link6` dominates. The doc row is struck and rewritten rather than deleted, so the correct-for-its-data reasoning stays visible.

### Safety argument

The envelope gets **tighter**, so containment is argued, not assumed:

- **Definitional, not fitted** — the DOP is the intersection of 26 tangent halfspaces `u·x ≤ h_mesh(u)`; the hull is `conv(mesh vertices)`.
- `generate_tight_geometry check` reports **mesh-outside-DOP `+0.000000000 mm`** on all seven links of **both** manifests.
- Hulls are 152/152/102 vertices — same class as the shipped `link2`/`link5`/`link7`, well inside `kMaxTightHullVertices`. `link1`'s 1588 stays stage-1 only.
- Both manifests are `hal.real: null`, so the **sim-margin** benchmark applies, where the staged path is **1.04–1.09× faster** than the box path. The 0.02 m real-HAL table would need re-measuring before either gains a real HAL.

**Still owed, and not faked here: a safety-WG reviewer and a hazard-log entry**, which `test_the_manifest_actually_declares_tight_geometry` names as the price of widening this set. This PR provides the measurement half.

## What I found and did NOT fix

- **`test_kernel_latency_soak.py` never publishes `OccupancyVoxels`** and runs a synthetic envelope with no collision geometry, so it *cannot reach the narrow phase* — its pass is vacuous for any hull change. The kernel's dominant cost path has no latency surface. Same class as the #183 vacuous Nav2 test. Tracked in `PLAN.md`.
- **Voxel resolution is struck as a lever**: `occupancy` is a dense `uint8[]` and the per-link window is `O(1/res³)`, so halving the cell is an **8× check cost**. 15 mm ≈ 26.7 ms against a 33 ms budget with a 2.80 MB message; 12.5 mm ≈ 46 ms. 20 mm fits but buys 4.4 mm of a 20.1 mm excess.
- **The real next lever**: 51 of 70 payload stops are against anonymous `voxel_` cells whose certified nearest body is a **static kitchen fixture MuJoCo already knows exactly** — `counter_1_right` ×25, `fridgesidebyside_main` ×9, `counter_1_left` ×8. Generalising #200's declared-target geometry to the fixture the payload is near is the measured next step, and is the survey's own §9 point 1.

## How tested

`test_collision_tight_geometry.py` (15), `test_collision_params.py` (9), `test_collision_geometry_contracts.py`, `test_all_manifests_validate.py` (126), `test_round_power.py`, `test_robocasa_sim_targets.py` — all pass, with the declared set, stage-2 count, CSR vertex counts, hull overhangs and box/DOP excesses **re-pinned to measured values**. The DOP-excess method was validated by reproducing all four previously pinned links exactly (53.27→25.69, 46.83→23.01, 45.20→18.99, 28.25→12.97).

`test_kernel_fridge_layout_pin_start_state.py` passes **5/5 against a real kitchen and a real kernel**, so the layout-47 pin, the layout-30 ordering, the genuinely-colliding pose and the all-occupied control all still hold under the tighter geometry. `ruff check tools/ tests/` clean.

Three defects in the ceiling harness were fixed before its number was trustworthy — a worker-killing `TimeoutExpired` that left the arms scene-confounded, a shared sidecar reaped mid-run, and policy-free runs needing exclusion (14 of 102). All recorded in `PLAN.md` §7 and the ledger.

https://claude.ai/code/session_01MTE4eEfxW8FvvBnPZL6otg


AdrianLlopart:
## The hazard-log entry this PR owed now exists

[OpenRAL/management#33](https://github.com/OpenRAL/management/pull/33) — **hazard-log Entry 026**, the record CLAUDE.md §3 and `test_the_manifest_actually_declares_tight_geometry` both name as the price of widening the tight-geometry set.

It carries the containment argument (definitional; `+0.000000000 mm` verified on 7/7 links of both manifests), the falsification of Entry 018's original exclusion (that reasoning was measured on **start states** and was right about them — the carry phase is where `link6` dominates), the latency measurement (**p99 2.0 ms over 5,638 occupied cells**, ~15× under the 33 ms ceiling), and three open items — chiefly that the 0.02 m real-HAL margin is unmeasured for seven hulls, which matters only if either manifest ever gains a real HAL (both are `hal.real: null` today).

**Safety-WG sign-off is still PENDING and I have not marked it approved** — I am not a WG reviewer. This PR now has its measurement half *and* its record half; what remains is a human ruling.

The same PR also carries **ADR-0101**, proposed and not implemented: modelling the static fixtures the carried payload *passes*, which is where the remaining 51 of 70 payload stops are. Recorded before code because it crosses Layer 2 → Layer 6.

AdrianLlopart:
## Scope grew: a defect in the instrument that adjudicates every stop

Four commits added since the description above, from chasing ADR-0101's premise against the **live** map rather than certified mesh truth.

### `voxel_backing_record` was stopping at decoration and blaming it

`mj_ray` reports only the nearest strike and the probe took it — so a non-collidable shell in front of the collidable slab it wraps was the only thing it ever saw, and the cell was adjudicated `noncollidable_world` ("the map disagrees with the world") when a real surface sat millimetres behind it **in the same cell**.

**6 of the 8 usable backing records** in the 2026-09-06 battery were misattributed this way, naming `counter_1_right_group_top_visual` while the certified collision surface was ~16 mm away.

The misreading was plausible because the class docstring still carried its **pre-#180** justification — *"the depth synth strikes these too, so they can become occupancy"*. #180 made exactly those geoms transparent to the cast. Corrected in code and METHODS.

**How much it moves**, on the real layout-47 kitchen grid:

| | occupied cells |
| --- | ---: |
| before (solid-only, as #224 measured) | 5,638 |
| **after** | **7,427** |
| all decoration counted (#224's upper bracket) | 9,217 |

**+32 %**, landing between #224's own brackets exactly as it should. Diagnostics only (§1.4) — the shipped grid comes from OctoMap, not this probe, so no deployed behaviour moves. Mutation-checked; fridge test still 6/6 on the denser grid; narrow phase p99 **1.7 ms on 7,427 cells**.

### A correction I made to my own ledger entry, kept visible

I first wrote that the backing record was present on "only 8 of 91 stops". **Wrong** — it is on 82, the extra 74 via #177's *late* path. But 46 of those read `unbacked`, which looks like a dramatic phantom-cell finding and is instead precisely the defect `10ff989` describes (the late path dropped `grid_orientation_xyzw`, decoded the wrong cube, and reported `unbacked` with 27 rays and 0 hits). **`10ff989` landed 2026-09-05; none of this battery's commits carry it.**

Kept as a visible correction rather than a silent edit, because the wrong version was the more exciting finding — which is the specific way this ledger has been burned before.

### Also since the description

- **[management#33](https://github.com/OpenRAL/management/pull/33)** — hazard-log **Entry 026** (the record this PR owes) and **ADR-0101**, recorded before code per §3.
- **ADR-0101's "Unknown" closed with a measurement:** 48 of 51 payload-vs-`voxel_` stops (94 %) would be recovered, median 16.2 mm; the 3 that correctly still stop are real penetration. Minimum recovered clearance **0.1 mm** — the number that argues for its suppression-off first landing.

All three checks green.

AdrianLlopart:
Split, per CLAUDE.md §4.2.5 — 4 822 lines across 33 commits was over the ceiling, and it had stopped being one logical change.

- **#238** `fix/collision-instrument-repairs` — the four instrument repairs, two evidence producers, `refine_dop_to_budget`, the ceiling probe, the narrow-phase latency surface, PLAN.md and the ledger. Touches no manifest and neither `packages/openral_safety/` nor `cpp/`, so §1.4 applies and §3 does not. **This is the urgent one**: `master` has carried two of those four defects since 2026-09-05, and both corrupt the programme's primary measurement, so every round taken on `master` since then is unusable.
- **#239** `feat/216-tight-geometry-link3-4-6` — the three manifest envelopes, stacked on #238, safety-WG gated on hazard-log Entry 026.
- **Withdrawn** — `panda_link1`'s refined envelope. The generator routine lands in #238; the manifest lands nowhere. Three rounds on `spark`, same seeds, one commit apart, moved its stops by 0.0003 mm. Shipping it would have added 41.989 mm of overhang to the link-link adjudication budget for a measured nothing.

Closing in favour of those two. The branch `feat/216-tight-geometry-carry-phase` stays as the archive of the sequence, including the two refuted hypotheses.

PR #238, opened 2026-09-07
fix(sim,hal,evidence): repair the four instruments the collision programme measures with

AdrianLlopart:
## What changed

Four instrument repairs, two evidence producers, one generator routine, and the
ledger that records what they measured. **No shipped envelope changes** — this
branch touches no manifest, no `packages/openral_safety/`, no `cpp/` (empty
`git diff --stat origin/master..HEAD -- robots/ packages/openral_safety/ cpp/`),
so CLAUDE.md §1.4 applies and §3 does not.

**The four repairs** — every one of them was *inflating* apparent severity, none
ever hid a contact:

| repair | what it was doing |
| --- | --- |
| `fix(sim)` DDS scope in `validation_matrix.py` | the harness could not discover the action server it had just launched — `ros2 action list --no-daemon` cannot see advertised actions, and the launch env was never scoped to `ROS_DOMAIN_ID=77` |
| `fix(evidence)` `nearest_any` inversion | a *permitted* adjacent-link overlap was reaching `nearest_any`, stamping every stop `real-contact`. Self-limiting — it can only turn a false positive into an apparent contact, never the reverse, so nothing was wrongly cleared |
| `fix(hal)` backing probe vs decoration | the probe stopped at a visual-only geom and blamed it for the stop |
| `fix(hal)` coincident shell + robot-only-cell sweep | the probe re-cast past a slab exactly coincident with its shell (how RoboCasa builds every counter top), and could not see world geometry in a cell whose only ray hit was the robot. Fixed with an AABB overlap sweep consulted only when the rays find nothing solid |

**Two evidence producers** — `tools/adr0101_recovery.py` (ADR-0101's recovery
rate, with an honest denominator and fixture-not-link attribution) and
`tools/stop_excess.py` (certified gap minus reported depth minus the voxel
half-diagonal — the geometry headroom left in a stop).

**One generator routine** — `refine_dop_to_budget`. It changes no manifest; see
"the one thing that is not here" below.

Plus the ceiling probe (`tools/_ceiling_probe.py`), the narrow-phase latency
surface (the kernel's dominant cost path had none — its soak test never
publishes `OccupancyVoxels` and so is vacuous for any hull change), `PLAN.md`
and `docs/reference/collision-validation-evidence.md`.

## Why now

`master` has carried two of these four defects since 2026-09-05. Both corrupt
the programme's primary measurement, so **every round taken on `master` since
then is unusable** and nothing downstream can be measured until this lands. That
is the whole reason this is a separate branch from the manifest work it was
developed alongside.

## The one thing that is not here

`panda_link1`'s refined envelope. The routine that builds it lands; the manifest
does not. It was generated, shipped to `spark`, and put under three rounds —
same seeds, same scene, one commit apart. It moved link1's stops by **0.0003 mm**.

The prediction failed because it read a support-census deficit against the
shipped *box* and attributed all of it to the hull, when the 26-DOP had already
collected that recovery (53.27 → 25.69 mm). Shipping it would have added
41.989 mm of overhang to the link-link adjudication budget and changed a
safety-critical link's certified envelope for a measured nothing. The refutation
is written into the routine's own docstring, into `docs/methods/10-tools.md` and
into the ledger, so the next reader cannot re-derive the expectation that failed.

## Size — §4.2.5 pre-approval requested

3 324 insertions, over the 800-line ceiling. Requesting maintainer pre-approval
rather than splitting further; the breakdown is why:

| | lines |
| --- | ---: |
| `PLAN.md` + evidence ledger | 1 261 |
| new standalone tools (4 files) | 808 |
| tests | 693 |
| production code (`sim_sensor_bridge.py`, `validation_matrix.py`) | 243 |

The production surface is 243 lines across two files, in five independent
one-function fixes. I tried splitting the four repairs into their own branch and
abandoned it: their "how tested" *is* the ledger, and the ledger's entries are
chronological and cross-referencing, so the split fragmented the document that
is itself part of the deliverable.

## How tested

- `pytest tests/unit` — **5 337 passed, 74 skipped**.
- `just lint` — ruff, `ruff format --check`, `mypy --strict` on 179 + 63 source
  files, all clean.
- `tools/refresh_methods_linenos.py --check` clean.
- Each repair carries the test that fails without it, against real fixtures
  (§1.11): a real RoboCasa counter with a coincident twin chunk, a real captured
  `run_deploy.log` snapshot line, the real robosuite link meshes.
- `refine_dop_to_budget` gets a direct test rather than a manifest assertion,
  precisely because no manifest exercises it: mesh inside the result, result
  inside the DOP, strictly less enclosed volume than the DOP it refines.

## Follow-on

`feat/216-tight-geometry-link3-4-6` stacks on this branch and carries the
`tight_geometry` manifests for `panda_link3`/`link4`/`link6`. That one is
safety-WG gated (§3) on hazard-log Entry 026 and should be reviewed separately.

Supersedes the instrument half of #235.


AdrianLlopart:
Added one more finding after the split, because it changes a decision this PR's PLAN.md carries.

**The voxel-resolution strike (PLAN.md §5, lever 3) was the one lever struck on an estimate rather than a measurement, and the estimate was wrong by 32×.**

Measured on the real kernel binary under a real rasterised RoboCasa grid at four resolutions over the same volume, 200 chunks each, 200/200 returned at every point:

| resolution | grid cells | occupied | median | **measured p99** | *§5 estimate* |
| ---: | ---: | ---: | ---: | ---: | ---: |
| **25 mm (shipped)** | 82 368 | 9 891 | 0.104 ms | **0.517 ms** | *5.8 ms* |
| 20 mm | 158 400 | 17 321 | 0.166 ms | **0.597 ms** | *11.3 ms* |
| **15 mm** | 376 680 | 35 828 | 0.184 ms | **0.825 ms** | *26.7 ms* |
| 12.5 mm | 630 054 | 59 948 | 0.162 ms | **0.838 ms** | *46.1 ms* |

Two errors compounded. The 5.8 ms baseline came from the hull microbenchmark, not from a round trip under a real grid — there was no latency surface to take that number from until this PR builds one. And the cubic factor was applied to the wrong term: `collision.cpp:1541` opens its window loop with `if (grid.occupancy[idx] == 0) { continue; }`, so `O(1/res³)` lands on a branch-not-taken while the real work scales with *occupied* cells, which are a surface. Cells ×7.65, occupied ×6.06, p99 ×1.62.

The cap objection fails too — 15 mm is 376 680 cells, under the shipped `world_voxel_max_cells = 614 125`. §5's 2.8M figure was a whole-kitchen grid, not the arm-neighbourhood window the kernel scans.

**Un-struck but explicitly not actionable.** This measures the kernel *consuming* a grid, not `openral_octomap_bridge` *producing* one at a finer tree resolution — the other half of the cost, unmeasured. PLAN.md now says so, and says the bridge measurement is the next step on this lever rather than a config edit.

`_RES` in the fridge pin file is now `OPENRAL_FRIDGE_GRID_RES_M`-overridable so the sweep reproduces from the shipped test instead of a probe duplicating it. Default unchanged.

Still no manifest, `packages/openral_safety/` or `cpp/` change on this branch.

Issue #227, opened 2026-09-05
fix(sim): deploy sim sets no DDS scope — a sim round joined a live robot's ROS graph


AdrianLlopart:
## Problem

`openral deploy sim` sets no DDS scope. It inherits the environment, and `ROS_DOMAIN_ID` is normally **unset — domain 0, with multicast discovery across the subnet**. On any network carrying a robot host, a *simulation* round therefore joins the *robot's* ROS graph.

This is not hypothetical. It happened on 2026-09-05 and silently destroyed a measurement.

## What happened

A validation-matrix A/B ran on `spark` (no hardware attached). `thor`, on the same LAN, was running a live bimanual OpenArm — `ros2_control`, `/joint_state_broadcaster`, `/openarm_{left,right}_hardware_interface`, `/safety_filter`, `/command_mux`, `/chunk_executor`, plus cameras. Neither host pins a domain, so both sat on domain 0 and discovered each other.

The sim read the robot's joints:

```
/joint_states  →  Publisher count: 2      # the sim's HAL AND thor's joint_state_broadcaster

ROSConfigError: state_contract.layout='rc365' binds gripper_qpos_joints=['panda_gripper'],
which the robot does not publish. It publishes ['openarm_left_finger_joint1',
'openarm_left_joint1', ... 'openarm_right_joint7'].
```

Every round bucketed `deadline-no-grasp` at ~50 s in **both** arms of the A/B — a result that looks like a policy failure and is nothing of the sort. Before #230 landed a typed error this surfaced only as `KeyError: 'panda_gripper'` from inside a layout file, relabelled by the runner's catch-all as a deadline.

Pinning `ROS_DOMAIN_ID=77` made the graph empty before launch, and the rounds went from **50 s / 133 steps** to **~4 min / 371-1744 steps**. Ten rounds then ran clean.

### The traffic went both ways

The sim also *published* `/joint_states` (with `panda_*` names) into the robot's graph, which had 3 subscribers on that topic. Nothing actuated — thor's stack consumes its own topic names and had no subscriber on `/openral/candidate_action`, `/openral/safe_action` or `/cmd_vel`, verified after the fact — but **that is a property of thor's topic naming, not a guarantee the design provides.** A sim publishing an `Action` onto a topic a physical robot subscribes to is one name collision away.

### A second contamination vector on the same code path

[...]

PR #251, opened 2026-09-08
fix(sim,hal,evidence): repair the four instruments the collision programme measures with, and measure the resolution lever end to end

AdrianLlopart:
## What

Slice **A** of the three-way split recorded in `0c2202d`: the instrument repairs, the evidence producers, and the resolution-lever measurements. **Touches no manifest, and neither `packages/openral_safety/` nor `cpp/`** — verified by an empty `git diff --stat` against both — so CLAUDE.md §1.4 applies and §3 does not. Slice B (`feat/216-tight-geometry-link3-4-6`, the three manifest envelopes) stays safety-WG gated on hazard-log Entry 026; slice C (`panda_link1`'s envelope) was withdrawn and lands nowhere.

## Why

**`master` has carried two measurement-corrupting defects since 2026-09-05, and every validation round taken on it since is unusable.**

1. **The harness could not see the graph it launched.** `_launch_env` never applied the sim DDS scope (#227/#231 confined the deploy to domain 77; the harness polled domain 0), *and* `ros2 action list --no-daemon` cannot discover an advertised action at all. Every scene of every round reported `harness-error` beside a healthy graph.
2. **The adjudicator stamped every stop `real-contact`.** #220 folded the new link-vs-link probe into `nearest_any`, whose first rule is "any probed pair ≤ 0 m → `real-contact`". Adjacent robot links overlap permanently and are ACM-permitted, so that rule became vacuously true. Three of four stops in the day's battery flip to `within-quantization` on re-derivation. The error runs one way — it manufactures real contacts, never clears one — so nothing was wrongly passed as safe.

Two more, both diagnostics-only: the backing probe could not see a slab **coincident** with its visual shell (RoboCasa builds every counter top that way), nor world geometry in a cell whose only solid backing is the robot.

Nothing downstream can be measured until this lands. That is the whole argument for splitting rather than waiting for the WG on one PR.

## What it establishes

With repaired instruments, **71 % of stops are of a physically clear robot** (5 of 7; +0.67 … +24.86 mm true clearance) — reproducing the #204 battery's 85-of-91 on an independent battery, a different commit, and a fixed adjudicator. The headline number has not moved.

It also measures the 25 → 15 mm resolution lever end to end, after correcting a strike that was **mine and wrong by 32×**:

| term | measured |
| --- | --- |
| kernel consuming a 15 mm grid | p99 **0.825 ms** (estimated 26.7 ms) |
| bridge producing one | **1.60 ms** vs a 100 ms period |
| the wire | **+15 ms median / +50–60 ms p99** staleness |
| carry-phase arm speed | 0.051 m/s median, 0.265 m/s max |

Net positive in three of four corners, and free for the whole start-state class (43 % of stops, arm speed exactly 0). **No manifest edit here** — shipping 15 mm is a conservatism *reduction* and needs its own WG ruling.

`_world_voxel_max_cells` now derives the cap from the resolution. That is a **fail-open fix**: a hand-kept 614 125 against a 15 mm grid's 2 803 221 cells makes the kernel reject every grid it is sent, which reads as "no world".

## How tested

```
109 passed   tests/unit/{test_stop_excess,test_adr0101_recovery,test_validation_matrix,test_sim_estop_voxel_backing}.py
5334 passed, 75 skipped   tests/unit (full tier)
12 passed   openral_octomap_bridge test_octree_to_grid (incl. the new resolution sweep)
1 passed    tests/sim/.../test_kernel_fridge_layout_pin_start_state.py -k latency
```

`ruff` clean, `mypy --strict` clean over `tools/` (65 files) and `openral_cli`, `tools/refresh_methods_linenos.py --check` exit 0.

Every behavioural fix is mutation-checked: restoring `+ link_link` fails the adjudicator test, forcing the backing sweep off fails the coincident-geometry test, flipping `>` to `>=` fails the contact-boundary test, and pointing the fixture reader at `nearest_link_link_pairs` fails the attribution test.

## Checklist

- [x] Conventional commit title; description has "What changed", "Why", "How tested"
- [x] Schemas: unchanged
- [x] Layer boundary crossed → none; no `packages/openral_safety/` or `cpp/` change
- [x] Tests: unit + sim; no new mocks/stubs (CLAUDE.md §1.11 — every fixture is a recorded artifact with provenance in `SOURCE.txt`)
- [x] `docs/methods/` updated for `stop_excess`, `adr0101_recovery`, `stop_ee_speed`, `voxel_transport_probe`; `--check` clean
- [x] Docs updated in the same PR (`PLAN.md`, `collision-validation-evidence.md`, standing caveats 10 and 11)
- [x] Pre-existing errors in separate prior `fix(...)` commits (`e8e0731` batches the hull-overhang query)
- [x] Repo state map — no module added, renamed, removed, or status-flipped
- [x] `just lint` passes; `mypy --strict` clean; no new bare `type: ignore`; no new `try/except: pass`
- [x] No new safety-disabling flag
- [x] Performance budgets met; no PII

https://claude.ai/code/session_01MTE4eEfxW8FvvBnPZL6otg


AdrianLlopart:
## The ADR-0101 re-derivation this PR's instruments made possible has now run

Thirty-six `utensil` rounds on `spark` at `448818c4` — the sha carrying both backing-probe fixes, a condition no `q-laptop` round had ever met, which is why this was blocked until the repairs in this PR existed.

| | offline (certified truth) | live map, re-derived |
| --- | ---: | ---: |
| payload-vs-`voxel_` stops | 51 | **28** |
| recovered | 48 (**94 %**) | **24 (86 %)** |
| **median recovered clearance** | **16.2 mm** | **8.56 mm** |
| minimum recovered clearance | 0.1 mm | 1.84 mm |

**The rate survives** — Fisher two-sided **p = 0.237**, no evidence the two differ. **The clearance distribution does not:** the median recovered stop sits at half the real air the offline analysis implied, which shrinks the margin ADR-0101's own modelling error has to fit inside. That is a number the safety-WG should have before ruling.

Two side effects that bear on this PR directly:

* **The stop decomposition now rests on n = 28, not n = 4** — payload **−9.93 mm** beyond the voxel term, link **+4.28 mm**. The exhaustion conclusion in `PLAN.md` §5 holds harder than when it was drawn.
* **The carry-phase yield entry in this PR is wrong, and I have corrected it in-branch.** It measured 18–27 % yield and 40–64 % `deadline-no-grasp` across 26 `q-laptop` rounds and read that as a property of the policy. Spark produced **zero** `deadline-no-grasp` in 36 rounds and 78 % payload stops. Same policy, different host — it was a 420 s deadline meeting a machine at load 19 with a shared GPU. `deadline-no-grasp` should be read as a load symptom first, which also means the ceiling battery's own policy-free exclusions deserve re-reading in that light.

Nothing here changes the diff under review; both corrections are documentation commits on this branch.

AdrianLlopart:
Superseded by #252.

This branch was cut before #238 and #239 landed, and had drifted into a 45-commit / 25-file diff that mostly **duplicates work already merged through review**. Rather than force-push a rebase over it, #252 rebuilds the same work on current `master` as one commit touching 10 files.

**Nothing is lost.** I compared every file against master (AST-compared where only prose differed) and kept master's reviewed version wherever the two agreed functionally. What was genuinely new here — the resolution-lever measurements, `voxel_transport_probe.py`, `stop_ee_speed.py`, the octree sweep test, `octomap_resolution_env`, `_world_voxel_max_cells`, and the PLAN/evidence entries — all carried over to #252 and is verified present there.

The pre-rebase tip is preserved locally at `556cb02` if anything needs recovering.

Issue #217, opened 2026-09-05
fix(sim): bisect the task-success drop — 5/20 on 2026-08-26 against 1/20 on 2026-09-04, in both arms


AdrianLlopart:
From the 2026-09-04 post-#200 battery (#209). Recorded there as **a flag for a separate bisect, not a verdict** — this is that bisect.

## The observation

| battery | commit | successes |
| --- | --- | ---: |
| 2026-08-26, five rounds | `7cb2376` | **5/20** |
| 2026-09-04, Path A **off** (shipped default) | `aceeca6` | **1/20** |
| 2026-09-04, Path A **on** (0.05 m band) | `aceeca6` | **1/20** |

Same seed, same host (`q-laptop`, RTX 5070 Laptop), same four scenes, same harness.

## Why this is not already answered

**Neither path predicts it and neither arm explains it.** Path A moves the robot 45 % further before the run ends and nearly eliminates the link-vs-world stop class (6 → 1), yet success is 1/20 with the band on *and* off. Whatever moved is upstream of the mechanism the battery was built to measure.

**It is also not established.** Fisher exact on 5/20 vs 1/20 is **p = 0.18** — not separable. Pooling both post arms (2/40 vs 5/20) reaches **p = 0.036**, but pooling two different configurations is arguably not legitimate. So this issue is "find out whether there is anything here", not "fix the regression".

**And the rollouts are not paired.** They diverge at the first stop, so five runs of a scene are five independent trajectories with no shared geometry; `task_success_steps` differs on every one. A naive A/B on two commits will not settle it either.

## What makes it worth doing anyway

29 commits separate the two batteries, several of which are *deliberately* more conservative and could plausibly cost completions:

- **#202** retired the `panda_link5`↔`panda_link7` ACM exemption — a pair that was previously never checked now stops the run. It **did** stop two runs in this battery.
- **#204** converted the support-contact witness to a certified instrument whose failure direction is a *missing* exemption.
- **#201** lowered `base_scan` `range_min_m`.
- **#186** made Nav2 base-only.

Each is correct on its own terms. The question is what they cost together, and nobody has measured that.

## Proposed approach

1. **Power first.** Decide the round count from the effect size worth detecting before running anything — 20 runs cannot resolve this, which is the lesson of the n=1 reading the 2026-08-26 battery had to take back.
[...]

Issue #191, opened 2026-08-31
fix(safety): remove the proven-real panda_link5<->panda_link7 ACM exemption


AdrianLlopart:
## Problem

#169 proved by **branch-and-bound over the pose box** (not sampling) that `panda_link5`↔`panda_link7` genuinely interpenetrates in **914 of 14,641 poses** — i.e. the committed allowed-collision matrix was **exempting a real, checkable pair** — and it refused a sampling threshold on principle. But #169 changed no manifest byte: `robots/panda_mobile/robot.yaml:751` still carries `- [panda_link5, panda_link7]` in `allowed_collision_pairs`, explicitly "under protest" (survey §2.2, ACM theme: *"#169 changed no manifest, so the unsafe exemption is still in the manifests"*).

A disabled real check on the self-collision path is exactly the state CLAUDE.md's safety posture exists to prevent. The reason it was exempted is known: OBB envelopes false-positive on **79.6 % of the interpenetration band** for this pair, so enabling it naively would spam stops.

## Proposed solution

Two options, in preference order once measured:

- **Option A (minimal):** delete the pair from `allowed_collision_pairs` and accept the OBB false-positive rate; measure what it actually costs on the live scenes (the 79.6 % figure is over the interpenetration band, not over visited configurations — the visited-configuration rate is the unknown that decides this).
- **Option B (if A's rate bites):** add `link5`/`link7` to `tight_geometry` so the staged 26-DOP → exact-hull path does this pair at hull fidelity — #166 measured exact hulls at 0.00 mm excess and 1.04–1.11× *faster* than the OBB path for link1/link2. Note the standing gate (survey §2.2): don't extend `tight_geometry` without a new measurement — a *self-pair* enable is precisely that measurement's subject (the map term doesn't apply to self-collision), so run the numbers first rather than treating the gate as a blanket ban.

Either way: the `allowed_collision_pairs` block is documented as `source: srdf`; after a manual edit that provenance line must record the override and cite #169. Safety-WG review applies (manifest change affecting the kernel's self-collision check).

## Verification

- #169's gtest flips from documenting the exemption ("pinned under protest") to **enforcing the check** for the pair.
- Five-round battery across the four validation-matrix scenes: quantify any new stop class attributable to link5↔link7 (this number decides A vs B); zero regressions in the other stop classes.
[...]

PR #252, opened 2026-09-09
feat(sim,evidence): the 25 → 15 mm world-voxel lever, measured end to end — and it is null

AdrianLlopart:
> **Status: draft, parked pending #266.** The measurement this PR set out to make is **done** and it came back **null**. The tooling is worth keeping and the evidence is worth landing, but the resolution question itself should be **re-asked after the payload-geometry investigation (#266)**, because that is what the null turned out to be about. Not ready to merge as a recommendation either way until then.

## What changed

- `OPENRAL_OCTOMAP_RESOLUTION_M` on `deploy_e2e.launch.py`, validated at launch and recorded per round, so a round that changed the resolution can never afterwards be mistaken for one that did not (7 tests).
- `tools/resolution_ab.sh` + `tools/resolution_ab_report.py` — the A/B and its offline report, with INTEGRITY (observed `resolution_m` per arm) and PAIRING (per-round separation) attestations.
- `tools/voxel_transport_probe.py`, `tools/stop_ee_speed.py`, octree-to-grid C++ tests, harness fixes in `validation_matrix.py`.
- The evidence entry for the result, and `PLAN.md` §5 updated to strike the lever.

## Why (the plan) — and what actually happened

The endpoint was deliberately **not** completion rate: against the predicted 2.7 % → 10.8 % that needs 200 runs/arm for 80 % power (~33 h), and at the 40/arm a battery gives, power is 0.11. So the primary was the paired, continuous quantity the mechanism predicts — per stop, the over-approximation, whose half-diagonal goes 21.65 mm → 12.99 mm, i.e. a **−8.66 mm shift**.

**Three batteries were needed.** The first two were voided by instrument defects found along the way, both now fixed and both with their own evidence entries:

1. every round leaked an `octomap_server_node` + `octomap_voxel_bridge` pair whose Fast-DDS lock file starved the *next* run on that domain to zero action chunks (fixed in #265, merged);
2. the runner claimed to interleave the arms and didn't — `xargs -P` holds a slot for all `ROUNDS` rounds — and then, once genuinely concurrent, two deploy graphs (~4.9 GB each on a 15.4 GB host with a 9.1 GB baseline) exhausted swap and drove load averages to 225.

The third ran clean: **80 rounds, 79 valid, 0 bond teardowns**, both arms verified against their own observed `resolution_m`, every round paired at a median separation of **0 s**.

## The result

Comparing only rounds where both arms produced an adjudicable stop **on the same party** — mixing parties compares different collision events, not resolutions:

| | |
| --- | ---: |
| same-party paired shift | **+5.9 mm** |
| 95 % CI (bootstrap, 20 000 resamples) | **[−5.7, +17.7]** |
| predicted | **−8.66 mm** — *outside* the interval, 2.6 se away |
| sign test | 11 neg / 12 pos, `p = 1.000` |

Wrong sign, not a weak confirmation. **Do not quote the report's own PRIMARY line** (`−1.6 mm`): it pools pairs where the 25 mm arm stopped on `panda_link7` (excess 77–89 mm) against pairs where the 15 mm arm stopped on the payload.

## Why it is null — and what to do instead

The carried payload is lowered to a **local AABB**; no hull path exists for payloads at all. Over this battery's 424 samples its box reaches a median **50.78 mm** (max 88.22 mm) beyond the mesh, against a half-diagonal of 21.65 mm at 25 mm and 12.99 mm at 15 mm. **The box is 2.3–3.9× the entire quantisation term** — shrinking the cell removes 8.66 mm and leaves ~50 mm beside it. Filed as **#266**, which is the lever this PR's line in `PLAN.md` §5 used to be.

15 mm also makes the concentration *worse* (payload stops 79 % → 97 % of all stops) and costs the grid 0.6 → 2.8 MB republished at 10 Hz — 28 MB/s of DDS against 6.

**Phase, reconciled with `PLAN.md` §5.** Two orthogonal axes, both real. On **phase** (was a place declaration live at the stop) it is **33 placing / 20 carrying / 6 pre-grasp** — which is what §5's earlier *11 placing / 7 carrying* census measured, and it was right. On **obstacle identity** it is **9 hit the declared target / 44 hit surrounding scenery**. Crossed over the 53 payload stops:

| | hit the declared target | hit scenery |
| --- | ---: | ---: |
| **placing** (declaration live) | 5 | **28** |
| **carrying** (in transit) | 4 | 16 |

The **28** is the headline — the payload approaching its declared drop point and clipping the occupancy beside it, exactly the pattern #259 names, now measured at scale rather than at n=11. `utensil` is 20/20 carrying; the other three scenes are placing. **0** stops occur while reaching for an ungrasped object, so picking is untouched. Lever mapping: **#259 reaches the 33 placing stops and nothing else; #266 reaches all 53 regardless of phase.**

## How tested

7 tests on the resolution knob; `tests/unit/test_resolution_ab_pairing.py` drives the real runner through a seam and asserts no two rounds are ever live at once, that the arms alternate, that each keeps its own ROS domain, and that the sidecar boot wait never falls inside an arm pair; `tests/unit/test_resolution_ab_report_pairing.py` covers back-to-back lanes, alternating rounds, and data with no `started_at`. `ruff` and `mypy --strict` clean; method markers checked.

## Standing caveat

n = 16 same-party pairs, sd 22.2 mm. This excludes an effect the size of the predicted −8.66 mm; it could **not** resolve a −3 mm one. The claim is "the predicted effect is not there", not "resolution does nothing" — which is why this is parked rather than closed.

Closes nothing on its own. Refs #253 (ruling withdrawn — measured), #266 (the lever that replaces it), #259, #264.


AdrianLlopart:
## Rebased onto `master` (past #257 / #258) — 2026-09-10

`master` had rewritten the same `PLAN.md` §5 this branch measures, so the branch stopped merging cleanly. Rebased; the five original commits are intact, plus one unification commit on top (`0efeb03`). `--force-with-lease` pinned to the pre-rebase tip `fcfb57d`, so nothing anyone else pushed was overwritten.

### Conflicts, and how each was resolved

| hunk | kept | dropped, and why |
| --- | --- | --- |
| PLAN §5 lever 3 | **this branch's** cost record — kernel 0.825 ms, rasterize 1.60 ms, wire +55 ms p99, carry speed n=5, the `_world_voxel_max_cells` correction | master's *"octree→grid conversion still unmeasured"* — it was wrong the moment this branch landed |
| PLAN "drop `baguette`" | **master's** withdrawal | this branch's `[x]` — `baguette` scored 1/3 with the gate off on the fixed harness, so the 0/11 behind the strike was starved runs |
| PLAN "Decide #217" | **this branch's** `[x] closed COMPLETED` | master's open item — it is a fact |
| `10-tools.md` | master's two new detector entries **plus** this branch's `octomap_resolution_env` | nothing; markers refreshed *after* the hooks ran |

`validation_matrix.py` and the evidence ledger auto-merged; `mypy --strict`, `ruff`, and the 110 tests across both branches' suites pass on the merged result.

### What the unification commit adds that neither side had

1. **A reach bound on this branch's own lever-3 item.** "Recovers 8.7 mm in every stop class" is true of the *term*, not the *stops*. Tracing the 20 adjudicable gate-ON stops to certified truth (#258): only 8 have an excess a 25 mm cell can account for, 3 of those already sit under 12.99 mm, so **25 → 15 mm plausibly converts 5 of 20**. The other 9 (excess 23–85 mm) are #254's population and no resolution reaches them. That's now written next to the "Actionable, with two caveats" paragraph rather than contradicting it from another file.

2. **A finding the gate-OFF arm exposed.** With the world check *off*, seven runs still carry a kernel stop, and every adjudicable one is a **hull-fidelity false positive on the exact pair the kernel named**: `link5↔link7` three times at −27…−32 mm reported vs **+1.4…+4.9 mm** certified; `link1↔link7` at +87.9 mm; payload↔`link1` at +25/+69 mm. No voxel in the path. They sit *inside* the 62.5 % ceiling and depress it, and `link5↔link7` is the pair #191 retired the exemption for. No issue exists yet.

3. **"What to test next"**, ordered by what is unblocked and what each answer changes. The 15 mm A/B this PR lists as missing is item 3, with a falsifiable prediction attached (~5 of 20 stops convert; fewer than 3 means the class-B bound is wrong).

The PR description above predates all of this; the numbers in it still hold, they are just no longer the whole picture.

AdrianLlopart:
Superseded by #268 and closing.

This PR's head is `experiment/voxel-resolution-15mm-v2`, an earlier line that **does not carry the A/B harness** (`tools/resolution_ab.sh`, `tools/resolution_ab_report.py`) or the null result — that work lives on `experiment/253-15mm-live-ab`, which was created from this branch and continues it. This PR was also `CONFLICTING` against current master.

#268 is that continuation, rebased onto master with #266 merged, pushed to a fresh ref so nothing here is rewritten. This branch and its history stay intact as the record.

**It also carries a correction this PR's title gets wrong.** "— and it is null" is true of what was measured and misleading now: the null held *because the payload was a box*. The terms add (`stop when true clearance ≤ payload overhang + cell reach`), so taking 8.66 mm off a ~50 mm sum changes nothing; after #266 took the payload overhang to ~9–10 mm, the same 8.66 mm is worth **89 %** of the remaining false payload stops. Measured by replay on the real kernel — see #268.

PR #257, opened 2026-09-09
fix(nav2,harness): a 4 s Nav2 bond timeout was scoring dead graphs as policy failures

### LINKED ISSUE #256: The ceiling battery's absolute completion rates were measured under contention — 31.1 % / 2.3 % are lower bounds
## The number

The 2026-09-07 ceiling battery is the measurement that reopened the collision programme — with the world-voxel gate off, **31.1 % completion (14/45)** against **2.3 % (1/43)** with it on, Fisher `p = 3.5e-04`. The **29 points of completion** derived from that difference is what justifies spending anything further on geometry at all.

## Why the absolute rates are suspect

That battery ran **both arms simultaneously** on one host, explicitly so that contention and scene composition would be shared. It excluded **14 of 102 runs** as policy-free by reading each run's goal log for `ROSConfigError` / sidecar-exit.

What it did **not** exclude is `deadline-no-grasp` — runs where the policy never grasped inside the 420 s deadline. Those counted as valid runs that failed to complete.

On 2026-09-09 that outcome turned out to be **largely a host-load artifact**:

| host | rounds | `deadline-no-grasp` | payload stops |
| --- | ---: | ---: | ---: |
| `q-laptop` (load ~19, shared GPU) | 26 | **40–64 %** | 27 % |
| `spark` (idle) | 36 | **0** | **78 %** |

Same policy, same scenes, same seeds-family. A 420 s deadline meeting a loaded machine produces `deadline-no-grasp`; an idle one does not.

## What this does and does not threaten

**The contrast is probably safe.** Both arms ran simultaneously under the same contention, so load deflates both roughly equally and the paired design absorbs it. The direction and the significance almost certainly hold.

**The absolute rates almost certainly do not.** `31.1 %` and `2.3 %` are **lower bounds** measured under load. If a meaningful share of the non-completions in either arm were load-induced `deadline-no-grasp` rather than collision-induced stops, then:

- the true ceiling is **higher** than 31.1 %, and
- **the 29-point figure — the headline that justifies the remaining geometry work — is itself a lower bound**, possibly a loose one.

## What would settle it

Re-run the ceiling A/B on `spark`, which is now known to produce zero `deadline-no-grasp` on these scenes. The harness, the gate-off path and the scene set all exist; this is a scheduling question, not an engineering one. Recording per-round load alongside the verdict would also stop this recurring.

## Related

`deadline-no-grasp` should be read as a **load symptom first** and a policy property second. That reading is now recorded in the 2026-09-09 correction entry of `docs/reference/collision-validation-evidence.md` (PR #252), which also withdraws an earlier "18–27 % carry-phase yield" finding that had made the same mistake.

AdrianLlopart:
Closes #256.

## What changed

#256 asked whether the ceiling battery's absolute rates (31.1 % / 2.3 %) were deflated by host load, and proposed re-running on an idle host to find out. They are deflated — but the cause is a fixable defect, not scheduling, and the battery's own artifacts answer it without a re-run.

**The reconstruction reproduces the published headline**, which is what makes the rest trustworthy: re-deriving each run's outcome from `run_deploy.log` alone gives 14 completions gate-off and 1 gate-on, against the published 14/45 and 1/43. Bucketing the non-completions is what the original never did — **61 % of the gate-off denominator never grasped**.

`run_goal.log` carries a throughput measure nobody had read, `latest_chunk`:

| | n | median chunks | range |
| --- | ---: | ---: | --- |
| completed | 15 | **544** | 233–769 |
| `deadline-no-grasp` | 37 | **16** | 0–128 |

No overlap. Healthy runs measure 1.15–1.8 chunks/s; a 420 s run delivering 16 chunks ran at 0.04/s and was never given a trial.

**Not contention, and not the shared XR-1 sidecar.** Both were the obvious suspects and both are wrong: a serialised ZMQ queue degrades `1/N`, but the distribution is bimodal, and once concurrency is measured at a *fixed instant* rather than over each run's own window (a fast run is a short run — the naive measure is circular) it predicts nothing. Runs with one other worker active are starved 100 % of the time.

**It was Nav2.** `lifecycle_manager_navigation` tears down the *entire* nav stack when a managed server misses its bond heartbeat, and the default timeout is **4 s**:

| server that lost its bond | runs |
| --- | ---: |
| `controller_server` | 25 |
| `planner_server` | 4 |
| `collision_monitor` | 2 |
| `behavior_server` / `smoother_server` | 1 each |

Across all 89 valid runs an early bond loss splits them perfectly — **31 flagged, not one `completed` or `estop` among them**. The teardown prints no traceback and exits non-zero nowhere; the graph stays up, goes inert, burns its deadline, and the harness scored the corpse as the policy failing to grasp.

`packages/openral_nav2_bringup/config/nav2_panda_mobile.yaml:499` already documents this cascade for a *different* trigger that was fixed. The timeout itself was never raised.

## Why this matters beyond the battery

The contrast was never in danger — both arms were hit about equally, which is exactly what the paired design absorbs, and Fisher *strengthens* (`p = 2.4e-04 → 1.3e-05`). The **absolutes** were badly deflated, and with them the headline: **the 29-point figure is a lower bound, and a loose one** — the real gap is nearer 60–70 points. The programme's justification is stronger than the ledger recorded.

## The fix

1. **`BOND_TIMEOUT_S = 30.0`.** It could not go in the params file: upstream `navigation_launch.py` hands `lifecycle_manager_navigation` only `{autostart, node_names}` and never the params file, so a block there is silently ignored — the trap this had to step around. Applied as a scoped `SetParameter` and **verified on the live node** (`ros2 param get` reports 30.0). This is a liveness timeout on the navigation stack, **not a safety check** — the E-stop path is `openral_safety_kernel` and is untouched — and it is raised, not disabled, so a server that really dies is still caught.
2. **`validation_matrix._nav2_bond_teardown`** — a run voided this way is a `harness-error`, not a policy outcome. The threshold sits in a measured 99 s empty gap (worst dead run `t0+100.8 s`, earliest healthy teardown `t0+199.6 s`) and is **declined outright** on a `_deploy_excerpt.log`, which begins mid-run: a missed teardown leaves today's behaviour, while a false one would silently drop a real result out of the denominator.
3. **Per-round load recording** — the probe now records host `loadavg`, delivered chunk count and any bond teardown alongside each verdict. This is #256's "record per-round load and stop this recurring", and the reason this took a log excavation rather than a query.
4. **`ceiling_battery.sh`** gains `WORKERS` (default 2, was a hard 8) and interleaves its worklist by scene, so the two live lanes are one scene's off and on arm and the arms stay paired.

## How tested

- `tests/unit/test_validation_matrix_nav2_bond.py` — 4 tests on **real deploy logs** from the 2026-09-06 battery: a torn-down run is voided, two completed runs are spared (including the tightest real case, bond lost at 199.6 s), and a mid-run excerpt is never anchored.
- Detector replayed over all 111 archived runs: 31 flagged, 0 healthy.
- `BOND_TIMEOUT_S` verified on a live `lifecycle_manager_navigation`.
- Existing suite: 83 validation-matrix tests, 114 CLI/nav2 tests pass. `mypy --strict` and `ruff` clean. `refresh_methods_linenos.py --check` clean.
- **The battery is re-running now** on a fixed harness (2 workers). Early rounds: 142 and 199 chunks, zero bond teardowns. Results will be added to the evidence entry when it lands.

## Docs

Evidence ledger entry + supersede notice on the 2026-09-07 entry; PLAN.md rates marked floors; the contributing doc's outcome table and harness-error list; repo state map; methods (`_ceiling_probe.py` and `ceiling_battery.sh` had no entries at all). A claim in the `adr0101-live-*` entry that made the same mistake — "roughly half of what looks like collision-programme failure is the policy" — is **withdrawn**.

`deadline-no-grasp` is not a policy property. It is the bucket every silent instrument failure falls into, because it is defined by absence, and absence is what a dead graph produces.


AdrianLlopart:
## Update: a second silent mode, and the correction no longer needs a threshold

Asking what the bond detector *missed* found the rest of it. Of the 16 starved runs it did not explain, **12 never brought their graph up at all**:

```
RuntimeError: transition 'configure' on '/openral_hal_panda_mobile'
did not advance the FSM within 300.0s
```

That one is *loud* — it raises — and it still landed in `deadline-no-grasp`. Which is the whole point: the bucket is defined by **absence**, and a graph that never started produces absence just as a torn-down one does.

`_lifecycle_never_came_up` needs no threshold and no clock, only the observation that a completed lifecycle transition is a precondition for a run existing at all. It appears in 12 of the 89 valid runs, every one `deadline-no-grasp`, and in **no** run that completed its task or was stopped by the kernel. With the bond detector it voids **43 of the 47** starved runs and **not one healthy run**.

### The correction, restated without a hand-chosen cutoff

Re-adjudicated by the shipped detectors alone, against the archived artifacts — reproducible from committed code:

| | valid runs | completed | rate |
| --- | ---: | ---: | ---: |
| world-voxel gate **OFF** | 21 | 14 | **66.7 %** |
| world-voxel gate **ON** | 25 | 1 | **4.0 %** |

**A 62.7-point gap** against the published 29. Fisher strengthens from `p = 2.4e-04` to **`5.8e-06`**.

It voids 25 runs from the off arm and 18 from the on arm — not arm-biased — and removes **no** completion from either. The four starved runs still unexplained are left in, which biases the result *down*.

This replaces the throughput sweep as the headline correction; the sweep is kept in the evidence entry as the independent check that found the mode, and it agrees (63.6 % / 4.5 % at 0.20 chunks/s).

Docs, methods and the contributing outcome table updated for the seventh harness-error mode. 85 unit tests pass; `mypy --strict`, `ruff` and `refresh_methods_linenos --check` clean.

The re-run on the fixed harness is still going — 14 of 80 runs, **zero bond teardowns** so far.

AdrianLlopart:
## The re-run landed: 62.5 % vs 2.7 %, a 59.8-point gap

Measured on the fixed harness — `q-laptop`, `WORKERS=2`, 10 rounds x 4 scenes x 2 arms, interleaved by scene. **80 runs, 69 valid, and zero Nav2 bond teardowns**, against ~35 % of runs in the 2026-09-06 battery. That is the fix, confirmed on live data.

| scene | gate **OFF** | gate **ON** | p |
| --- | ---: | ---: | ---: |
| `utensil` | **9/10 (90 %)** | 0/10 (0 %) | 0.0002 |
| `fridge` | **7/10 (70 %)** | 0/10 (0 %) | 0.002 |
| `sink_cup` | 3/9 (33 %) | 1/10 (10 %) | 0.25 |
| `baguette` | 1/3 (33 %) | 0/7 (0 %) | 0.30 |
| **pooled** | **20/32 = 62.5 %** | **1/37 = 2.7 %** | **3.1e-08** |

### Three numbers for the same quantity

| | OFF | ON | gap |
| --- | ---: | ---: | ---: |
| as published, 2026-09-07 | 31.1 % | 2.3 % | 29 pts |
| the archive re-adjudicated (this PR) | 66.7 % | 4.0 % | 62.7 pts |
| **this run, measured** | **62.5 %** | **2.7 %** | **59.8 pts** |

The re-adjudication and the fresh measurement agree within 3 points by **independent routes** — one corrects a denominator on old artifacts, the other never had the defect. The published pair is the outlier. The 29-point headline was a floor, and the true gap is about twice it.

### One conclusion withdrawn

The 2026-09-07 entry argued **`baguette` should leave the collision scorecard** — 0/11 with the gate off, so policy-bound and unable to report on collision work. It scored **1/3** here. That 0/11 was partly starved runs. Weak, not dead; it stays on the scorecard until it has ten valid rounds.

### Caveats, stated rather than buried

A concurrent GPU job on the host during the opening lanes cost **11 runs** (`baguette-off` 7, `baguette-on` 3, `sink_cup-off` 1; the `fridge`, `utensil` and `sink_cup-on` lanes lost none). The probe **refuses to score** these rather than guessing, so they do not corrupt the rates — but `baguette-off` rests on 3 valid rounds instead of 10, and the pooled OFF figure is weighted toward the strong scenes. **The evidence entry says to read the per-scene table, not the pooled rate**, until a top-up lands; the deficit and its commands are recorded beside the data. The arms are also unbalanced (32 valid OFF vs 37 ON — the gate stops runs early, so fewer are lost).

CI was green on the previous head and this push is docs-only.

AdrianLlopart:
## Correction: the rates I posted here were floors again (#262)

The **62.5 % / 2.7 % / 59.8 points** I posted above are wrong, in the same direction and for the same reason as the correction they were themselves reporting.

That battery's goal logs carry **6 `ConnectivityException`** (the TF tree still two disjoint trees at dispatch) and **17 `ROSConfigError: XR-1 expected camera …`** across 78 logs. Each returns in ~0.4 s having delivered **zero action chunks**, and each was scored as the policy failing — the graph survives, prints `sim.task_success_final` at teardown, and reads as an ordinary non-completion. **21 of the 69 valid runs delivered zero chunks.**

| | as posted | policy-free | **corrected** |
| --- | ---: | ---: | ---: |
| gate **OFF** | 20/32 = 62.5 % | 7 | **20/25 = 80.0 %** |
| gate **ON** | 1/37 = 2.7 % | 14 | **1/23 = 4.3 %** |
| gap | 59.8 pts | | **75.7 pts**, Fisher `p = 5.5e-08` |

Per scene, once the policy actually acts:

| scene | OFF | ON | p |
| --- | ---: | ---: | ---: |
| `utensil` | **9/9 (100 %)** | 0/8 | 0.00002 |
| `fridge` | **7/7 (100 %)** | 0/5 | 0.0013 |
| `sink_cup` | 3/6 (50 %) | 1/5 (20 %) | 0.35 |
| `baguette` | 1/3 (33 %) | 0/5 | 0.38 |

**Cause:** the harness waits for the action server, sleeps a fixed **5 s**, and dispatches. Enough on an idle host, not enough under contention. Fixed by re-dispatching — bounded and recorded — while the only thing wrong is that the graph is not assembled yet.

**This is the third time the same mistake has landed on this ledger and it was mine each time**: a bucket defined by absence read as a policy property. The 2026-09-10 entry itself says *"read `latest_chunk` on every record before believing any rate"*, and then published a rate without doing so.

**Unaffected:** the stop census, the A/B/C class split, and everything in #260 — **0 of the 22 gate-ON stops occurred in a zero-chunk run**, since a stop requires the arm to have moved. So the lever analysis stands; only the completion rates move, and they move the same way as before — they were floors.

Issue #221, opened 2026-09-05
fix(evidence): measure each tight hull's overhang past its source mesh, so a hull-adjudicated self stop has a budget


AdrianLlopart:
The actionable half of #215, separated out because it needs **no real-time code**.

## What is still unadjudicable after #220

#220 gave the validation matrix a link-vs-link budget, keyed on the kernel's own `depth_is_box_bound` (#213):

| `depth_is_box_bound` | kernel measured | budget | outcome |
| --- | --- | --- | --- |
| `1` | the OBB's bound | `2 × corner_slop` | adjudicable |
| `0` | the exact hulls | **none** | `unadjudicated` |

The second row is the gap. A stop where the kernel judged two stage-2 hulls **separated but inside the margin** reports a hull-to-hull distance, and the probe reports mesh-to-mesh. The admissible gap between them is the amount each hull reaches *beyond* the mesh it was built from — and that number is measured nowhere. #220 deliberately refuses to charge the box budget instead, because doing so would forgive a real overlap by up to `2 × corner_slop`, which is **172 mm** on `panda_mobile`.

## Why this is not #215

#215 asked for the *overlapping* branch's depth to be measured (EPA). Its motivating case — the 2026-09-04 `panda_link5`↔`panda_link7` stop — is now fully handled: flagged as a box bound by #213, charged the box budget and scored `real-contact` by #220. EPA would tighten a number that is already conservative, disclosed and scorable, at the cost of a containment-preserving polytope search on an allocation-free real-time path. An attempt is written up on #215; it did not converge and was backed out.

This issue is the remaining unadjudicable class, and it is **offline**.

## Proposed solution

`tools/generate_tight_geometry.py` already holds both the source collision mesh and the convex hull it emits for every link. At generation time, compute per link:

```
hull_overhang_m = max over hull surface of  distance(point, source mesh)
```

— the furthest any point of the hull sits outside the mesh. For a convex mesh it is 0; for `panda_link5`'s 152-vertex hull it is whatever the concavities cost. Ship it in the manifest beside `hull_vertices_m` (`TightCollisionGeometry`, `python/core/src/openral_core/schemas.py`), so it is a **declared, testable property of the geometry the kernel actually loads** rather than something re-derived per round.

Then `estop_ground_truth_snapshot` (`python/hal/src/openral_hal/sim_sensor_bridge.py`) fills the field #220 left as `None`:

```
[...]

PR #261, opened 2026-09-10
fix(harness,hal): charge the box budget when a self-pair link has no stage-2 hull

### LINKED ISSUE #260: A self-pair naming a link with no stage-2 hull is permanently `unadjudicated` — the box budget is never charged
## The defect

`hal_admissible_gap_m` decides which budget a link↔link self stop gets by reading `depth_is_box_bound`:

```python
if stop.depth_is_box_bound:
    return link_block["admissible_gap_box_m"]
return _link_link_hull_gap_m(budget, stop)     # ← treats a clear flag as "hull fidelity"
```

But the kernel's flag has a narrower meaning than that. `hull_hull_distance` sets it when **a hull refinement was attempted and fell back to the OBB**, and deliberately leaves it clear when no refinement happened at all:

```cpp
if (a.vertices == nullptr || a.n_vertices <= 0 || b.vertices == nullptr || b.n_vertices <= 0) {
    return fallback;  // one side stops at stage 1; the OBB bound stands
}
```

That behaviour is intentional and tested — `SelfCollisionHull.TheFlagIsClearedWhenNoHullRefinementHappened`: *"a plain box verdict must not advertise itself as a hull measurement that fell back. Fail-toward-honest: unset means 'this is the ordinary box path', which is what it is."*

So a clear flag is **two** different facts, and the adjudicator collapses them into one. For the stage-1-only case it charges a hull budget, then `_link_link_hull_gap_m` finds the link has no measured `hull_overhang_m`, returns `None`, and the stop scores **`unadjudicated`**.

**The HAL's own published rule already states the correct condition** — `adjudication_budget.link_link.rule`:

> …or — **when both links ship stage-2 tight geometry** (#191/#202) — exact hull to exact hull… the admissible gap is `corner_slop(a) + corner_slop(b)` for the **BOX** case, or `hull_overhang_m(a) + hull_overhang_m(b)` for the **HULL** case

The implementation never checked "when both links ship stage-2 tight geometry".

## Why it bites now

**`panda_link1` ships no stage-2 hull by decision.** Its refined envelope was withdrawn because the battery measured it moving link1's own stops by **0.0003 mm** (`PLAN.md` Path C). The manifest carries a `tight_geometry` block with **zero** `hull_vertices_m` and a null overhang — correct and deliberate.

The consequence was unintended: **every self-pair naming `panda_link1` is permanently unadjudicable.** In the 2026-09-10 ceiling battery, 2 of the 7 self stops:

| run | pair | depth | `depth_is_box_bound` | verdict |
| --- | --- | ---: | --- | --- |
| `fridge-off/r09` | `panda_link1` ↔ `panda_link6` | −2.8 mm | false | **`unadjudicated`** |
| `utensil-off/r01` | `panda_link1` ↔ `panda_link7` | −1.7 mm | false | **`unadjudicated`** |
| `sink_cup-off/r03,r05,r08` | `panda_link5` ↔ `panda_link7` | −27…−32 mm | **true** | `within-quantization` ✓ |

It is silent in the worst way: `unadjudicated` reads as *missing evidence*, not as *the scorer cannot see this case*. The evidence is complete; the scorer is blind to it.

**This nearly produced a false safety finding.** Reading those stops without a budget — comparing kernel depth against certified mesh distance directly — makes them look like gross false positives, which is precisely what `hal_admissible_gap_m`'s docstring warns of: *"recomputing only the voxel term understates it fourfold… turning correct stops into `false-positive`."* An earlier draft of the `PLAN.md` entry said exactly that and had to be withdrawn.

## Not a stop-behaviour bug

Nothing here changes what the kernel stops on. `panda_link5`↔`panda_link7` at −27…−32 mm is the kernel finding its **hulls genuinely overlapping** and reporting the loose OBB bound because it runs no expanding-polytope step — `collision.cpp` names this exact pair in its own comment ("−31.97 mm for a ~1.5 mm hull interpenetration"). It discloses that, and the adjudicator charges the 176.4 mm box budget correctly. That path is working as designed.

## Fix

Diagnostics-only, both halves:

1. **HAL** publishes `has_stage2_hull` per link in `collision_model_slop`, so a consumer can tell "no hull" from "hull whose overhang was never measured".
2. **Adjudicator**: when either named link has no stage-2 hull, charge `link_link.admissible_gap_box_m` — what the kernel actually measured with. A hull that exists but has no measured overhang still leaves the pair unadjudicable, which is correct: there the kernel really did refine and nothing bounds the hull-to-mesh gap.

**Absence of the field must read as "unknown", never "no hull"** — the reverse would hand the box budget to genuine hull measurements on every historical round and turn correct stops into `within-quantization`. Only an explicit `False` is load-bearing.

## Verification

Replayed over the archived battery with the field set as the manifest actually declares it:

```
fridge-off/r09    link1<->link6   unadjudicated -> within-quantization  budget=0.17644
utensil-off/r01   link1<->link7   unadjudicated -> within-quantization  budget=0.17644
sink_cup-off/r03  link5<->link7   within-quantization -> within-quantization   (unchanged)
```

All seven self stops in the battery become scorable, and all are `within-quantization` — correct conservative behaviour throughout.

Artifacts: `openral-256/outputs/ceiling/2026-09-09-fixed/` on `q-laptop` (gitignored, host-local). Related: #256, #258.

AdrianLlopart:
Closes #260. Diagnostics-only — no stop is suppressed, delayed or altered.

## What changed

`hal_admissible_gap_m` chose a link↔link budget by reading `depth_is_box_bound`, treating a clear flag as *"the kernel measured at hull fidelity"*. The kernel's flag is narrower than that: `hull_hull_distance` sets it when a hull refinement **was attempted and fell back**, and deliberately leaves it clear when none was attempted — the stage-1-only path, where either side carries no hull vertices. That is intentional and already tested (`SelfCollisionHull.TheFlagIsClearedWhenNoHullRefinementHappened`).

So a clear flag is **two** facts and the adjudicator collapsed them. For the stage-1-only case it charged a hull budget, found no measured `hull_overhang_m`, and returned no budget at all.

The HAL's own published `link_link.rule` already states the right condition — the hull term applies *"when **both** links ship stage-2 tight geometry"*. The implementation never checked it.

- **HAL** publishes `has_stage2_hull` per link in `collision_model_slop`.
- **Adjudicator** charges `link_link.admissible_gap_box_m` when either named link has no hull. A hull that *exists* with no measured overhang still leaves the pair unadjudicable — correct, since there the kernel really did refine.
- `_lacks_stage2_hull` gates on an explicit `False` only. Absence reads as **unknown**, never "no hull" — the reverse would hand the box budget to genuine hull measurements on every historical round and turn correct stops into `within-quantization`.

## Why it bites

`panda_link1` ships no stage-2 hull **by decision** — #191 withdrew its refined envelope for moving link1's own stops by **0.0003 mm**. The unintended consequence: every self-pair naming `link1` scored `unadjudicated` forever. In the 2026-09-10 ceiling battery that is 2 of the 7 self stops, and it fails silently — `unadjudicated` reads as *missing evidence*, not as *a scorer blind to this case*. The evidence was complete.

**It nearly produced a false safety finding.** Read without a budget, those stops look like gross false positives — the exact error `hal_admissible_gap_m`'s docstring warns of (*"turning correct stops into `false-positive`"*). An earlier `PLAN.md` entry claimed precisely that and has been withdrawn in #252.

## How tested

`tests/unit/test_self_pair_box_budget.py` — five tests, all four branches plus the manifest premise:

| test | asserts |
| --- | --- |
| manifest premise | `panda_link1` still ships a stage-1-only block; every other link still has a hull |
| no stage-2 hull | box budget charged |
| hull, overhang unmeasured | still `None` — the kernel did refine |
| field absent (historical) | old behaviour preserved |
| two hulled links | #221 sum untouched |

Replayed over the archived battery with the field as the manifest actually declares it:

```
fridge-off/r09    link1<->link6   unadjudicated -> within-quantization  budget=0.17644
utensil-off/r01   link1<->link7   unadjudicated -> within-quantization  budget=0.17644
sink_cup-off/r03  link5<->link7   within-quantization -> within-quantization  (unchanged)
```

All seven self stops become scorable, and all are `within-quantization` — correct conservative behaviour throughout. 106 existing unit tests pass; `mypy --strict` and `ruff` clean; methods markers refreshed post-hook.

## Not in scope

`panda_link5`↔`panda_link7` reporting −27…−32 mm for a ~1.5 mm hull interpenetration is the kernel having no expanding-polytope step. It discloses that via `depth_is_box_bound` and is budgeted correctly; `collision.cpp` names the pair in its own comment. Tightening it would be an EPA step in the kernel, which this PR does not attempt.

## Open items at review time

- **Deferred, not unresolved** — whether `has_stage2_hull` should instead be a kernel-side `depth_source` disclosure (`hull` / `box_fallback` / `no_refinement`), putting the fact at its origin rather than reconstructing it from the manifest. Filed as #264 with both options written out: it touches the kernel's evidence surface (an IDL field, a `schema_version` question, a migration for every archived snapshot) and retires a deliberately-tested flag behaviour, so it is out of scope for a diagnostics-only PR. Nothing here is wrong under either answer — the question is where the fact lives.
- **Confirmed on a live round** (`cf4c982`). `utensil`, gate **on**, q-laptop, its own `ROS_DOMAIN_ID`: a kernel stop at `sim_time_s = 20.4`, `stop_class: attached_payload`, and its `adjudication_budget.collision_model_slop.links` reads `panda_link1` → `has_stage2_hull: false` / `hull_overhang_m: null` with every other kernel-checked link `true` and a measured overhang — the shape the manifest declares. Raw line committed as `docs/reference/data/estop-snapshot-has-stage2-hull-2026-09-10.jsonl`; the round is written up in `docs/reference/collision-validation-evidence.md`. Driving `hal_admissible_gap_m` with that same live budget block: `link1`↔`link6` and `link1`↔`link7` get **0.17644 m** (the box term, `None` before this PR), `link6`↔`link7` gets 0.000323 m (the #221 sum), and a `depth_is_box_bound` pair is unchanged — and the 176.44 mm equals the archived-battery replay figure to the digit. The producer is also pinned at unit tier by `b12b529`, against a real compiled `panda_mj_description` MjModel and the real `panda_mobile` manifest.
- **That seam is now pinned** (`82d7174`, its own `fix(test)` commit since the seam predates #260 — §1.15). `test_the_snapshot_publishes_the_slop_block_where_the_adjudicator_reads_it` asserts the block arrives at `adjudication_budget.collision_model_slop` — the path between the pinned producer and the pinned consumer, which nothing covered: a reader looking for a top-level `collision_model_slop` gets `None` and concludes the field was never published, which is what happened while reading the live round. Real upstream `panda_mj_description` MjModel, real `panda_mobile` manifest, real `estop_ground_truth_snapshot` — the live round's components minus the ROS graph. Mutation-verified: renaming the key inside the budget dict fails it with `KeyError: 'collision_model_slop'`. 19 tests pass in the file.


AdrianLlopart:
## Second open item closed — the producer is now proven against real meshes

The unit tests covered the manifest and the adjudicator, but nothing exercised the **HAL producer** that has to emit `has_stage2_hull`. That gap is closed in `b12b529`.

`test_collision_model_slop_is_tight_on_faces_and_loose_at_corners` already loads a real `panda_mj_description` MjModel *and* the real `panda_mobile` manifest to compute the corner-slop budget, so the assertion rides on those rather than adding a stand-in:

- `panda_link1` → `has_stage2_hull: False`, `hull_overhang_m: None`
- every other kernel-checked link → `True` with a measured overhang

38 tests pass across the three affected files. That leaves one item open on this draft — whether the fact belongs kernel-side as a distinct `depth_source` disclosure (hull / box-fallback / no-refinement) rather than being reconstructed from the manifest. That is a design question for review, not a gap in what is here.

Checklist status:
- [x] Sim-tier confirmation that a live round emits the new field
- [ ] Whether `has_stage2_hull` should instead be a kernel-side `depth_source` disclosure

AdrianLlopart:
## Correction to the comment above, and the draft items closed out

The comment above ticked **Sim-tier confirmation that a live round emits the new field**. That was too strong and the body now says so plainly. What `b12b529` proves is the **producer**, at unit tier, against real components — a real compiled `panda_mj_description` MjModel and the real `panda_mobile` manifest, not a stand-in. It is not a live round: no real `estop_ground_truth_snapshot` has been captured on this branch carrying the field end to end. Read that item as **owed**.

The second item is now **deferred, with both options written out** — #264. Whether `has_stage2_hull` should instead be a kernel-side `depth_source` disclosure (`hull` / `box_fallback` / `no_refinement`) touches the kernel's published evidence surface: an IDL field, a `schema_version` question, a migration for every archived snapshot, and the retirement of a deliberately-tested flag behaviour. Out of scope for a diagnostics-only PR, and nothing here is wrong under either answer — the question is where the fact lives, not what it says.

The program page in the management repo now carries the landing: **OpenRAL/management#39** — the P5 (#191) row (the link this bites is the one #191 chose to leave alone), the "Done so far" line, and a dated change-log entry that records both items above as open rather than done.

So the review question is just the one: is the manifest-side derivation the right shape to merge, with #264 tracking the kernel-side alternative?

AdrianLlopart:
## The live round is in — the field travels, and the budget it earns is the one the replay predicted

The last open item on this PR was sim-tier confirmation on a *live* round. Run now, on q-laptop (RTX 5070 Laptop 8 GB, host otherwise idle), one round per `ROS_DOMAIN_ID`. Landed as `cf4c982`.

**`utensil`, gate on** — kernel stop at `sim_time_s = 20.4`, `stop_class: attached_payload`, one `sim.estop_ground_truth_snapshot`. On the live wire, `adjudication_budget.collision_model_slop.links`:

| link | `has_stage2_hull` | `hull_overhang_m` |
| --- | --- | --- |
| `panda_link1` | **`false`** | `null` |
| `panda_link2`…`panda_link7` | `true` | 0.0000890 – 0.000485 |

Exactly what the manifest declares — `panda_link1` alone ships no stage-2 hull, because #191 withdrew its refined envelope for moving link1's own stops by 0.0003 mm. The raw line is committed as `docs/reference/data/estop-snapshot-has-stage2-hull-2026-09-10.jsonl`.

**The branch this PR changed, driven by that live record.** The stop itself is `attached_payload`, so its own adjudication does not take the link↔link path. Asking `hal_admissible_gap_m` what the *same live budget block* yields for a self pair:

| pair | budget | before #261 |
| --- | --- | --- |
| `link1`↔`link6` | **0.17644 m** (box term) | `None` |
| `link1`↔`link7` | **0.17644 m** (box term) | `None` |
| `link6`↔`link7` (both hulled) | 0.000323 m (#221 sum) | unchanged |
| `link5`↔`link7` (`depth_is_box_bound` set) | 0.17644 m | unchanged |

The 176.44 mm is the live `link_link.admissible_gap_box_m` and it matches the archived-battery replay to the digit — two independent routes to the same number.

**A gate-on round that completed the task, worth recording.** The first round attempted was `fridge`, gate on: 470 s, **753** action chunks, kernel never tripped, `ever_succeeded: true` — so it emitted no snapshot, which is why a second round was needed. `fridge`-on was **0/10** in the 2026-09-10 ceiling battery. One completion is not a rate, but it says those per-scene zeroes are small samples rather than impossibilities.

**And one thing the live run exposed that this PR is not fixing.** The field is nested at `adjudication_budget.collision_model_slop`, and nothing asserts that nesting — the producer test reads `collision_model_mesh_slop` directly, the adjudicator tests build the budget dict by hand. A first pass at reading the live line looked at a top-level key and wrongly reported the field absent. The plumbing predates #260 and is not wrong, so under §1.15 it is a separate change; it is named in the evidence doc so it is not rediscovered.

Program page updated to match: **OpenRAL/management#39** now records the live round instead of an owed item. The only open question left on this PR is #264 — kernel-side `depth_source` vs the manifest-side derivation.

AdrianLlopart:
### Correction to the host-state line above

I wrote "host otherwise idle" for the live round. That overstated it, so `b88d79c` states what was actually true.

The host **was** verified idle at launch — no `octomap` orphans, no live deploy graph, GPU at 15 MiB. Another session's resolution A/B battery then started a worker at ~20:27, during the first round; its own deploy graphs came up at 20:41, after both rounds here had finished (20:32:56 and 20:36:00), so **no two live deploy graphs ever overlapped** — though its XR-1 sidecar was loading on the same 8 GB GPU during the second round.

Nothing changes in the finding: whether a JSON field is present in a snapshot is not contention-sensitive, and the one timing-flavoured observation (a gate-on round completing the task) is *harder* under load, not easier. Corrected because the claim was wrong for part of the window. OpenRAL/management#39 carries the same qualification.

Issue #264, opened 2026-09-10
Should `has_stage2_hull` be a kernel-side `depth_source` disclosure instead of a manifest-side derivation?


AdrianLlopart:
Follow-up to #260 / PR #261, which is the reason the question exists rather than a gap in what landed there.

## The question

#261 made "this link ships no stage-2 hull" a fact the adjudicator can read, by having the HAL publish `has_stage2_hull` per link in `collision_model_slop` (derived from whether the manifest's `tight_geometry` block carries hull vertices). `tools/validation_matrix.py::_lacks_stage2_hull` then charges `link_link.admissible_gap_box_m` when either named link has no hull.

That reconstructs, on the *manifest* side, a fact the *kernel* already knows at the moment it measures. `hull_hull_distance` distinguishes three cases internally and reports only one bit of them:

| what happened in the kernel | `depth_is_box_bound` today |
| --- | --- |
| hull refinement ran and succeeded | clear |
| hull refinement was attempted and fell back | **set** |
| no refinement attempted (a side carries no hull vertices) | clear |

Rows 1 and 3 are indistinguishable in the evidence, which is exactly the collapse #260 was. #261 resolves it by re-deriving row 3 from the manifest instead of asking the kernel to say which of the three it did.

The alternative is a `depth_source` disclosure on the evidence record — `hull` / `box_fallback` / `no_refinement` — set by the code that actually took the branch.

## Why it might be the better shape

- The fact lands at its origin. A consumer reads what the kernel did, instead of inferring it from a second artifact that has to stay in sync with the manifest the kernel was built from.
- It is robust to the manifest and the running kernel disagreeing. `has_stage2_hull` describes the manifest as loaded by the HAL; `depth_source` would describe the measurement that produced the number being adjudicated.
- It subsumes `depth_is_box_bound`, which is a two-valued projection of a three-valued fact.

## Why it was not done in #261

- It touches the kernel's evidence surface — an IDL field in `packages/msgs/`, a `schema_version` question for on-disk snapshots, and a migration story for every archived `estop_ground_truth_snapshot`. #261 is diagnostics-only and adds no schema.
- `depth_is_box_bound`'s current behaviour is deliberate and pinned by `SelfCollisionHull.TheFlagIsClearedWhenNoHullRefinementHappened`; replacing it means retiring or rewriting that contract.
[...]

Issue #259, opened 2026-09-10
The place allowance is scoped to the declared target, so a payload clipping neighbouring occupancy at the drop point still E-stops (9 of 20)


AdrianLlopart:
## The finding

In the 2026-09-10 ceiling battery (#256 / #257 — 80 runs, fixed harness, zero Nav2 bond teardowns), the gate-ON arm produced **20 hard latching E-stops**. Classified by manipulation phase from each run's own `automatic sim attachment revision`, `support_witness_separated` and `place_declaration_armed` markers:

| phase | hard E-stops |
| --- | ---: |
| **PLACING** — a place declaration was armed before the stop | **9** |
| CARRYING — lifted clear of support, no declaration yet | 7 |
| grasped but not yet lifted | 1 |
| an arm link, not the payload | 3 |
| approach / before the grasp | **0** |

**Not one stop happens during the pick.** And in **all 9 placing stops the payload hit a world voxel that is not its declared place target** — the counter top beside the drop point, the fridge shelf next to it — while `place_allowance_active` was `False`.

## Why

`PlaceDeclaration` (`openral_core.schemas`) scopes to a single `target_id`, and its docstring is explicit that on its own it exempts nothing — it only lets a producer attest support contact against **that** target. The kernel's allowance therefore covers the declared surface and nothing else.

That is *correct as specified*. The problem is that placing a thing on a counter means bringing it within millimetres of the rest of the counter, and the rest of the counter is ordinary world occupancy carrying the full 21.65 mm cell inflation. The declaration exempts the square you aimed at, not the square centimetre next to it.

`place_allowance_active` is `False` on **all 20** stop lines, so the allowance is not merely too narrow here — it never armed at all on any stop in this battery.

## The nine stops

| scene | reported depth | **certified real gap** | nearest real surface |
| --- | ---: | ---: | --- |
| `fridge` | −11.7 mm | **−0.7 mm** | `fridgesidebyside_main_group_1_g25` |
| `sink_cup` | −11.2 mm | **−0.2 mm** | `island_island_group_top_right_0` |
| `baguette` | −1.1 mm | **−0.1 mm** | `counter_1_left_group_top_left_1` |
| `fridge` | −14.8 mm | +0.0 mm | `fridgesidebyside_main_group_1_g25` |
| `fridge` | −12.1 mm | +0.2 mm | `fridgesidebyside_main_group_1_g25` |
[...]

Issue #253, opened 2026-09-09
Safety-WG ruling needed: ship the world-voxel grid at 15 mm (a deliberate reduction in conservatism)


AdrianLlopart:
## What is being asked

Whether to change the sim world-voxel resolution from **25 mm to 15 mm**, which drops the kernel's own quantisation term from **21.65 mm to 12.99 mm**.

**This is a reduction in conservatism.** Every one of those 8.66 mm is protection the kernel has today and would give up. It is not a new mechanism, a new flag or a bypass — one parameter changes value.

Decision record: **hazard-log Entry 027** (`OpenRAL/management#36`). Measurements: PR #252, `PLAN.md` §5, and the 2026-09-08/09 entries in `docs/reference/collision-validation-evidence.md`.

## Why it is worth ruling on

It is the only lever that touches **every** stop class — payload, bare link, and start-state alike. ADR-0101 covers only a carried payload against modeled bodies; tight geometry covers only link envelopes.

The original strike on this lever was **wrong by 32×**, because it rested on an estimate rather than a measurement. All three cost terms are now measured on the real kernel, the real rasterizer, and real DDS:

| term | 25 mm (shipped) | 15 mm | budget |
| --- | ---: | ---: | --- |
| kernel narrow phase, p99 | 0.517 ms | **0.825 ms** | 33 ms @ 30 Hz |
| bridge rasterization | 0.88 ms | **1.60 ms** | 100 ms @ 10 Hz |
| publish→receive, p99 | 19–23 ms | **68–83 ms** | — |

**Compute is not the constraint; the map's age is.** Latency triples, and age is millimetres too. Against measured arm speed (carry-phase 0.051 m/s median / 0.265 m/s max; start-state exactly 0) the trade is net positive in three of four corners and free for the whole start-state class.

## The asymmetry the ruling turns on

The **8.66 mm gain is deterministic**. The **staleness cost is probabilistic** — it depends on arm speed and on transport tail latency. Those are different kinds of guarantee, and that is the substance of the decision rather than any single number.

## What does not exist, and may reasonably be required first

- [ ] **No live A/B of completion rate at 15 mm vs 25 mm.** Every figure above is a cost or a mechanism; none is evidence the change converts stops into completions. The harness and the `OPENRAL_OCTOMAP_RESOLUTION_M` override can both run it today.
[...]

PR #267, opened 2026-09-11
feat(safety): give attached payloads the stage-2 hull the links have

### LINKED ISSUE #266: Attached payloads are lowered to a local AABB, not a hull — a median 50.8 mm of the collision envelope is the box, and it is now the dominant stop cause
## Summary

Robot links got stage-2 hulls. **Attached payloads never did** — `extract_body_primitives` lowers a *mesh* geom to its **local AABB**, and there is no hull path for payloads at all. This is not a fallback after a failed refinement; it is the only lowering implemented for a carried mesh.

Measured across all 80 rounds of the 2026-09-10 resolution A/B (424 samples, 4 scenes):

| payload box corner slop | value |
| --- | ---: |
| median | **50.78 mm** |
| max | **88.22 mm** |

Identical in every scene. For scale, the world-voxel half-diagonal that #253 proposes to shrink is **21.65 mm** at 25 mm cells and **12.99 mm** at 15 mm. **The payload box contributes 2.3–3.9× the entire quantisation term.**

## Why this is now the top lever

The 25 mm → 15 mm A/B (#253) ran clean — 80 rounds, 79 valid, 0 bond teardowns, both arms verified at their declared resolution, pairing exact — and returned a **null**:

```
same-party paired shift   +5.9 mm   95% CI [-5.7, +17.7]
predicted                 -8.66 mm  → OUTSIDE the interval, 2.6 se away
sign test                 11 neg / 12 pos,  p = 1.000
```

It is null *because the voxel term is the small one*. Shrinking the cell removes 8.66 mm of envelope while the payload box leaves ~50 mm in place.

Stop composition confirms where the error now lives:

| | 25 mm | 15 mm |
| --- | ---: | ---: |
| stops that are payload-vs-world | **27/34 (79 %)** | **36/37 (97 %)** |

## The stops are mostly not real contacts

True certified clearance at the moment the kernel stopped, per payload stop:

| true gap at stop | 25 mm | 15 mm |
| --- | ---: | ---: |
| touching / penetrating (≤2 mm) | 4 | 7 |
| close (2–10 mm) | 6 | 9 |
| **genuinely clear (>10 mm)** | **9** | **16** |

Median true gap **9.6 mm** in both arms. So roughly half the payload stops fire while the carried object is a centimetre or more from anything — which is exactly what a 50 mm box bound predicts.

**The check must stay.** The ≤2 mm bucket contains real interpenetration (`-7.7`, `-1.7`, `-1.5`, `-1.3` mm) — the check catching a carried object already in contact. Deleting payload-vs-world would recover most of the ~76-point gate-off/gate-on ceiling gap and would also let those through. The fix is fidelity, not removal.

## Phase: this is a carrying problem

Classified by what the payload was geometrically nearest at each of the 59 stops:

- **53 of 59** stops have the payload **already grasped**; only 6 are arm-only.
- **0** stops occur while reaching for an ungrasped object — picking is not affected.
- Of the 53, only **9 hit the fixture they were dispatched to**. **44 hit transit scenery**: the counter being crossed, a cabinet door, the fridge body.

| scene | task | what it hit |
| --- | --- | --- |
| `utensil` | CounterToDrawer | counter top ×16, drawer door ×4 |
| `sink_cup` | CounterToSink | island counter ×15, **the sink itself ×1** |
| `fridge` | FridgeShelfToDrawer | fridge body ×12 |
| `baguette` | CounterToCabinet | counter ×5, cabinet ×5 |

Note `PLAN.md` §5 records an earlier census as *11 placing / 7 carrying*. That is a labelling difference, not a data conflict — PLAN counts "descending toward the drop point but clipping the counter beside it" as placing. Worth reconciling before either number is cited again.

## What to do

1. Give attached payloads the same stage-2 treatment the links have — a convex hull per primitive rather than a local AABB, with the same `MAX_TIGHT_HULL_VERTICES` ceiling and the same honest disclosure when it cannot be built.
2. The adjudication budget already measures the payload's share (`attached_payload_mesh_slop`), so the before/after is directly measurable on archived rounds with no new battery.
3. Interacts with #259 (place allowance scope) and #264 (`has_stage2_hull` as a kernel-side disclosure). Does **not** need the #253 ruling — it is orthogonal and larger.

Data: `outputs/resolution-ab/2026-09-10-serial` on q-laptop (gitignored); per-stop rows in its `report.json`.

Refs #253, #259, #264

AdrianLlopart:
Closes #266.

## What changed

`extract_body_primitives` lowers a carried **mesh** geom to its **local AABB**, and that is not a fallback after a failed refinement — it was the only lowering a carried mesh had. Robot links got the staged 26-DOP → exact-hull narrow phase in #166; payloads were listed as *not replaced*, "there is no measured motivation".

`check_attached_voxel_collision` now runs the same two stages for a payload primitive that arrives with a proved-contained refinement, folded as `max(box, staged)`.

| | robot link (#166) | carried payload (this PR) |
|---|---|---|
| producer | `tools/generate_tight_geometry.py`, offline, once per robot release | `_tight_geometry_from_points`, online, at every grasp |
| over the vertex budget | `refine_dop_to_budget` | stage 1 only — that search is not affordable per grasp |
| carrier | `RobotDescription` manifest | three additive `AttachedCollisionPrimitive` wire fields |
| proof | `validate_tight_geometry` at `on_configure`, **fail-closed** | `ingest_attached_objects` per message, **fail-down** |
| scope | arm-link↔world-voxel + self-collision | payload↔world-voxel only |

Both sides share `validate_tight_hull` (kernel) and `check_tight_geometry_fits_box` (schema), extracted rather than copied, so a payload hull is never held to a weaker standard than a link's.

## Why

The 25 → 15 mm resolution A/B (#253) ran clean — 80 rounds, 79 valid, 0 bond teardowns, both arms verified at their declared resolution, pairing exact — and returned a **null**:

```
same-party paired shift   +5.9 mm   95% CI [-5.7, +17.7]
predicted                 -8.66 mm  → OUTSIDE the interval, 2.6 se away
sign test                 11 neg / 12 pos,  p = 1.000
```

Null **because the voxel term is the small one**. Over the same 424 samples (4 scenes, identical in every scene) the payload box's corners stand a **median 50.78 mm** (max 88.22 mm) proud of the mesh — against a world-voxel half-diagonal of 21.65 mm at 25 mm and 12.99 mm at 15 mm. **2.3–3.9× the entire quantisation term.**

Payload-vs-world was **27/34 (79 %)** of stops at 25 mm and **36/37 (97 %)** at 15 mm, and roughly half of those fired with the payload >10 mm clear of anything.

**The check stays.** The ≤2 mm bucket holds real interpenetration (−7.7, −1.7, −1.5, −1.3 mm). The fix is fidelity, not removal.

## The safety case

```
payload mesh ⊆ conv(surface points) ⊆ 26-DOP ⊆ local AABB (the box)
```

Definitional at every link — tangent halfspaces, a convex hull of the points, and the points' own AABB grown by 1e-4 m. No optimiser tolerance anywhere, so the broad-phase window never moves. Every stage is a **lower bound** on the true clearance and the kernel takes the max, so:

> **The refinement can never stop where the box would not have.**

Two decisions worth a reviewer's attention:

- **Ingest fails DOWN, not closed.** A payload's refinement arrives on every world-state message from a live producer. Refusing the attachment would drop the payload's geometry — the *unsafe* direction. Only the refinement is dropped; the primitive is checked as the plain box, pre-#266 behaviour bit for bit. An over-budget hull is refused **whole**, never truncated.
- **ADR-0098's place-target guard keeps reading the shipped box.** That bound is calibrated against the box model the declaration is adjudicated on. Inside that branch this PR is bit-for-bit inert; outside it, it removes stops.

## What actually fires, measured

Re-lowering every free payload in the four A/B scenes at seed 1: a real RoboCasa collision mesh carries **749–3319 vertices and is already convex** (MuJoCo compiles them that way), so **three of eight** payloads are over `kMaxTightHullVertices` and ship the DOP alone — the same representation, for the same reason, that `panda_link1` ships. Stage 1 is the uncapped stage and is tangent along all 13 axes, which is where the corner slop lives. Box corner slop **13.56–23.00 mm**, worst-direction support excess **7.97–13.95 mm**, wire **0.4–21 kB** per payload, producer 0.37–0.44 → 0.51–1.10 ms per lowering.

No convex-hull solver runs: stage 2 is a support scan and the support of `conv(V)` is the support of `V`, so SciPy would return the input unchanged at 4–17 ms a call.

## How tested

- `AttachedVoxelTightGeometry.RefiningAPayloadOnlyEverRemovesStops` — the never-adds-a-stop property, over a sweep of single-cell grids through the shipped entry point.
- `AttachedVoxelTightGeometry.TheCornerCellIsExactlyWhatTheBoxWasInventing` / `.ARealContactStillStops`.
- `AttachedIngestTightGeometry.AnUnprovableRefinementIsDroppedNotObeyed` — escaping slab, hull outside its DOP, over-budget hull, and no refinement at all.
- `tests/unit/test_attached_payload_tight_geometry.py` — the containment chain on compiled MuJoCo models, the analytic geoms that correctly ship nothing, the cluster path, the schema refusals, the wire round-trip, and the `n_stage2_primitives` disclosure.
- Run locally: **229** kernel gtests, **50** octomap-bridge gtests, **5401** unit tests, **4** RoboCasa sim tests (`tests/sim/test_attached_object_multitask_robocasa.py`) — all passing. `ruff`, `ruff format`, `mypy --strict` and `clang-format` clean.

## Owed, stated plainly

- **The 50.78 mm figure is not reproduced here.** Its rounds are gitignored on q-laptop (`outputs/resolution-ab/2026-09-10-serial`). Re-scoring the archived rounds is the measurement that would size the benefit, and it needs no new battery.
- **No live round.** Nothing here shows the removed stops convert into completions.
- **A world-voxel payload adjudication budget measured against the refinement.** `attached_payload_mesh_slop`'s `corner_slop_m` is the box's and stays the right number for a *self* stop; charging it to a world-voxel payload stop over-budgets a refined primitive. Mitigated for now by publishing `n_stage2_primitives` (the payload-side analogue of #260's `has_stage2_hull`).

## Also in this branch

A prior `fix(safety)` commit clears pre-existing clang-format drift in five kernel translation units and one codespell hit from #202 — both were blocking the pre-commit gate (CLAUDE.md §1.15).

**This PR is over the 800-line pre-approval threshold** (CLAUDE.md §4.2.5) and is safety-touching: it needs a safety-WG reviewer, and hazard-log **Entry 028** is in OpenRAL/management#41 with sign-off deliberately unticked.

Refs #253, #259, #264

https://claude.ai/code/session_01VVrppMan2ts7WZe944fkTL


AdrianLlopart:
Pushed `7f4ddbb`: CI's `select-and-test` was red on `tests/hil/test_openarm_restock_deploy_preflight.py`, which is **pre-existing** — it reproduces unchanged on `origin/master` in a clean tree. `06710e7` moved the cell's `top` slot to the `ros2_image` backend (the ZED's rectified left image exists only because zed_wrapper computed it), and two preflight assertions kept reading `device` unconditionally. Each now asks what its own comment says it asks. Its own commit, per CLAUDE.md §1.15.

AdrianLlopart:
## Measured on the real kernel

The "no live round" gap is narrowed. Not a completion-rate round — but the recovery itself is now measured through the **shipped binary**: real `safety_kernel_node`, real `openral_msgs` IDL, a real dense occupancy grid, and a real RoboCasa payload's lowered primitives on a 1-DoF prismatic carriage, so the commanded joint value *is* the payload's position. Zero margins, no declaration, so the refusal boundary is purely geometric. Swept twice against byte-identical everything else: once with the wire's `tight_*` fields stripped, once with them.

| payload | stage | recovered travel: min / **median** / max |
|---|---|---|
| `PickPlaceCounterToCabinet` `obj_main` (1 primitive, 98-vertex hull) | 1 + 2 | 5.10 / **17.65** / 28.40 mm |
| `PickPlaceCounterToSink` `obj_main` (2 primitives, over budget → DOP only) | 1 only | 0.00 / **8.75** / 18.70 mm |

16 orientations each, rotated about the primitive's own centre so the sweep measures the shape and not the payload's internal offset; boundary by ascending scan to 0.1 mm.

- **Never negative, 32 of 32.** The refinement never stopped the carriage earlier than the box did — §10.3's soundness argument observed rather than argued, on the shipped binary.
- **The size is right.** #253's predicted gain was **8.66 mm** and it measured null. The stage-1-only payload here — the *field-typical* case, since a real collision mesh is over the vertex budget — gives back **8.75 mm median** through the same kernel. Stage 2 roughly doubles it.
- **0.00 mm happens and is correct.** A box is only loose at its corners, so a face-on approach has nothing to give back.

**What it is not:** one occupied cell is not a kitchen and a carriage is not a policy. It sizes the conservatism removed, not how often that conservatism was costing a task. Stop-rate and completion-rate stay owed (§10.7).

Written up in `docs/reference/collision-hull-narrow-phase.md` §10.5, including the two traps that cost a rebuild each: the occupancy grid has a **deadline**, and a sweep that outlives it turns every chunk into a `voxel_unavailable` refusal that reads exactly like a geometric one; and the kernel **latches** on refusal, so a reset is only evidence the latch is gone once a known-safe chunk is accepted again.

## Review fixes (`2238628`)

A review caught that three sites claimed a producer "cannot widen the kernel's envelope, only fail to narrow it". **That was false as written**, and it is now corrected everywhere including hazard-log Entry 028. The proof shows the refinement is inside the box, so the broad-phase window still visits every cell — but it cannot show the refinement bounds the payload's *real* geometry, because the kernel never sees a mesh. It is not a trust boundary this change adds (`half_extents` arrives on the same message under the same trust, checked only for finite-and-positive), but the claim had to say so rather than imply a verification that does not happen.

Also: `check_attached_voxel_collision` range-checked `hull_index` at the `TightPose` init but used it under a weaker condition. Not UB today — `TightPose` zero-initialises and `accept_tight` already bounds the index — but a zeroed `TightPose` reports clearance the geometry does not have, so the asymmetry fails **open**. One predicate now.

Behaviour-neutral: 229 kernel gtests unchanged, and the live sweep returns the same 5.10 / 17.65 / 28.40 mm after the fix.

## On the diff size

The 2700-line figure in the description is not 2700 lines of code: **1180** auto-generated JSON schemas (`just schema-export`), **590** tests, **304** docs, **52** IDL, and **681** lines of hand-written code. 681 is under CLAUDE.md §4.2.5's threshold — I flagged it off the wrong number.

AdrianLlopart:
## High-effort review found a real containment hole — `e9549f5`

The earlier low-effort pass missed it. **`solid ⊆ conv(sampled surface points)` is an equality for a POLYTOPE and runs the wrong way for anything curved.** `geom_surface_points` samples a sphere at its six axis poles, so the support of those points along the 26-DOP's first corner axis is `r/√3 = 0.577 r` while the sphere reaches `r`:

```
sphere r=50 mm, DOP axis (0.577, 0.577, 0.577)
  sampled support  28.868 mm
  TRUE support     50.000 mm
  DOP cuts         21.132 mm INSIDE the real sphere   <- larger than a 15 mm voxel's 12.99 mm
```

The kernel would then report the payload farther from an occupied cell than it is — a **missed stop**, the one direction this PR must never produce.

It reached only `_clustered_box_primitives` (payloads with more than `max_primitives` geoms), which fed every geom's samples in regardless of type. **Nothing downstream could catch it:** `check_tight_geometry_fits_box` and `validate_tight_hull` compare only the three box axes, and the ten diagonal slabs have no box bound to violate. The same path also let a geom with *no* enumerable surface (ellipsoid, heightfield, SDF, plane) grow the cluster's box through its AABB corners while contributing nothing to the DOP.

Closed by `geom_is_polytope`, gating both lowering paths, with **one ungroundable geom disqualifying its whole cluster** — a DOP is a single solid bounding all of them and cannot be sound for only some. Costless in practice: a payload lowered one geom at a time never reaches the cluster path, and the four A/B scenes cluster meshes only.

**The lesson worth more than the fix:** my containment test could not have caught this. It proved `mesh ⊆ DOP` using the same sampler the producer used, so an under-stated geom was under-stated identically on both sides and the assertion passed by construction. A containment proof checked against its own construction proves nothing. The test now carries a mesh+sphere+capsule+cylinder cluster fixture and asserts the 21.13 mm deficit against the sphere's *analytic* support.

### Also fixed from the same review

- **`validate_tight_hull` accepted 13 zero slabs** — finite, not inverted, inside any box — collapsing the payload to a *point* after which every cell reads clear. Reachable from a producer that resized the arrays and forgot to fill them, not only from a deliberate lie. Positive extent on the three box axes is now required; pinned by a gtest through the shipped entry point.
- **Over-budget hull refused the refinement whole**, throwing away a DOP that validates and leaving the two sides of the wire disagreeing about what "over budget" means. Keeps the slabs, drops the vertices — stage 1, what the producer emits for the same condition.
- **`decode_tight_geometry` reserved against an unbounded `float64[]`** inside the safety node's subscriber. Bounded before the reserve.
- **`push_back(pin)` copied** up to 123 kB of vertices per object per message in a node whose contract is allocation discipline. Moved.
- **`_tight_geometry_from_idl` raised where the kernel drops** — one bad producer message would keep the kernel running while killing the world-state callback decoding the same bytes. Fails down now.

Hazard-log Entry 028 records the hole and the lesson rather than quietly absorbing them (OpenRAL/management#41).

Behaviour-neutral on valid input: 229 kernel gtests, 50 octomap-bridge gtests, 5404 unit tests, 4 RoboCasa sim tests, and the live carriage sweep returns the identical 5.10 / 17.65 / 28.40 mm.

AdrianLlopart:
CI note, so a reviewer does not trip on it: the first run of `e9549f5` failed the `robocasa` lane on `tests/sim/test_so100_robosuite_lift.py::test_gripper_closes_during_close_phase` — jaw `range=0.285` against a `> 0.3` threshold. **Re-run on the identical commit: `robocasa | ran | 18 passed`.** A flake, not a regression, and not reachable from this PR: that file's only module-level import is `pytest`, it references `openral_hal` nowhere, and it measures MuJoCo physics from a scripted open/close policy. It also passes locally 14/14 and the same lane passed on `7f4ddbb` with the same test selection. The jaw did reach its closed limit (max `0.500` exactly); only the resting-open position differed (`0.216` vs the ~0.10 the assertion's comment expects), so the threshold is marginal on that runner. Flagging rather than touching it — loosening someone else's threshold on one failure is not a fix.

Issue #272, opened 2026-09-12
The payload-authored-cell census was an instrument artefact (closed): the coincident-shell sweep was suppressed by the carried payload


AdrianLlopart:
## The finding

Two independent lines of work now point at the same hole: **E-stops fire against world occupancy that a correct voxelisation of the same scene does not contain, and the evidence path cannot say why.**

### Line 1 — the ceiling battery's two wide-gap placing stops (#259)

| run | task | kernel depth | certified gap to real geometry |
| --- | --- | ---: | ---: |
| `sink_cup-on/r04` | `PickPlaceCounterToSink` | −5.91 mm | **+44.96 mm** |
| `baguette-on/r05` | `PickPlaceCounterToCabinet` | −8.08 mm | **+48.53 mm** |

The #268 replay swept both of those **same tasks, same scenes, same `obj_main` payload** pose-by-pose under the shipped `box@25` model:

- `CounterToCabinet` `box@25` clears from **29.9 mm** upward — it does not stop at 48.53 mm.
- `CounterToSink` `box@25` stops **0 of 11** poses in the 40–60 mm band — it does not stop at 44.96 mm.

So the model that ran in the battery, on a correctly built grid of that scene, does not produce these stops. Something put occupancy there that the scene's own geometry does not.

### Line 2 — the #254 re-derivation's residual

Re-deriving the ADR-0101 classes after #267/#269 collapsed class C from 8 to 2. That was the basis for closing #254 as not-planned — the modeled-fixtures lever lost its population. But the **largest residual outlier was 280.89 mm of excess**, which no cell of any resolution and no collision-model conservatism explains. It is not a conservatism artefact; it is occupancy with no known source.

Closing #254 deliberately left this half unaddressed. This is that half.

## Why it can't currently be investigated

`estop_ground_truth_snapshot.evidence_voxel_backing` is **`null` on all 20 stop snapshots** in the 2026-09-09 battery. That is the one field that would resolve a stop's voxel back to the sensor evidence that created it. Without it, the only available argument is the replay comparison above — an inference, and a cross-artifact one.

## Why this matters more than the levers it displaced

Every open item in this program so far has been a request to *give up conservatism*: #253 (finer cells), #254 (modeled fixtures), #259 (wider place allowance). Each was measured and each lost most of its population once the geometry work landed.

[...]

PR #273, opened 2026-09-12
fix(hal): make the world-voxel stop record say what it means — and retire #272's premise

AdrianLlopart:
Makes the world-voxel stop record say what it means. Three changes to the same evidence path, and the last one **retires the issue this branch was opened for**.

## 1. `fix(hal)`: a carried payload must not suppress the coincident-shell sweep

`voxel_backing_record` takes a second look when the ray fans find no collidable **world** geometry — the sweep that finds a collidable slab behind a decoration shell, i.e. every RoboCasa counter and shelf. The condition excluded `robot_body_ids` alone, and **an attached payload is in neither the world nor the robot**. So the carried object suppressed the sweep exactly as a world geom would, while satisfying none of the reasoning that makes suppression safe.

False `attached_payload` verdict on the *normal* case: a payload resting on its support, both in one cell.

| `fridge` gate-ON, 6 rounds an arm | stops | `attached_payload` | `solid_world` |
| --- | ---: | ---: | ---: |
| before | 6 | **3** | 3 |
| after | 4 | **0** | 4 |

The baseline was self-contradictory without any battery: two of its three `attached_payload` stops had certified true gaps of **−2.30 and −4.27 mm** — the payload *inside* real geometry, in a cell reportedly holding only the payload.

Fisher on 3/6 vs 0/4 is `p ≈ 0.2`, and these are different stops under policy noise, so the comparison alone would not carry it. The mechanism is pinned directly against compiled MuJoCo in `test_a_payload_in_the_cell_must_not_hide_the_world_surface_behind_it`, which fails before and passes after.

## 2. `feat(hal)`: freeze the occupied-cell set at the grasp

`preattach_verdict` / `occupied_cell_keys` — a stop can now say whether its cell predates its own grasp. Measured live: `cells=7912, resolution_m=0.015, truncated=false`.

Two things the tests pinned that my first implementation got wrong: keys must be `floor(centre/res)` (with an origin on a multiple of the resolution, *every* cell centre sits on a rounding boundary), and they must name a **world** cell rather than a voxel index (the base-frame origin slides as the robot drives).

**`preexisting` is not a finding on its own** — a world-backed cell is always pre-existing. Only the conjunction with an `attached_payload` verdict means the payload put the cell there. Written into the docstring so it cannot be quoted the wrong way.

## 3. `cube_rot_world` on every backing record

Three attempts at an offline paired re-adjudication failed, and the last failure is a fact about the record, not the harness: only the cube's **centre** was stored. At 25 mm a geom centred ~44 mm out reaches inside on one orientation and not another. Without the axes, a classifier change can only be compared across *different* stops — the exact weak inference this PR had to work around.

## What this does to #272

**Its premise does not survive.** "6 of 20 stops trip on a cell with no world geometry" was substantially this misclassification. The payload leak is not established. `preexisting: true` still holds but means little once the cell legitimately contains the shelf — that is an ordinary correct stop near a support surface.

#272 is being closed on this.

## 4. Retraction propagated to every citation

The 2026-09-09 "6 of 20" figure measured the classifier. Every in-tree citation now says so rather than leaving a reader to find the correction three sections later: the evidence page's census row is struck through with the retraction inline, and `docs/methods/01-hal.md` and the live integration test name it where they used to cite the count.

`PLAN.md`'s "roughly 6 of the 20 stops" is deliberately untouched — that is the 8.5–18 mm clearance band cleared by 25 → 15 mm, a different claim that happens to share the number.

## Dropped before merge

`OPENRAL_OCTOMAP_ATTACH_SWEEP_PADDING_M` was added chasing a padding-regression theory that turned out to be a detour — the clearing withholds those cells deliberately, and once the cell legitimately holds the shelf, withholding is correct. The commit is rebased out rather than left in as unused instrumentation for a hypothesis that did not survive.

## Verification

Unit 26 pass; full unit tier 5454; live-ROS integration passes; `ruff`, `mypy --strict -p openral_hal`, `refresh_methods_linenos.py --check` clean. 12 live deploy rounds across two arms. Battery artifacts are host-local and not committed.

https://claude.ai/code/session_01VVrppMan2ts7WZe944fkTL


AdrianLlopart:
## Battery-tested, partially — and here is exactly which part

Asked to validate this on the battery before merge rather than on unit tests alone. Did that. It splits into two halves, both proven on real infrastructure, and one combination this host cannot produce.

### 1. The freeze fires at a real masking attach, in production — proven

One `fridge` / gate-ON round from this branch, real deploy graph, real XR-1:

```
automatic sim attachment revision 1: ['sim:obj_main']
sim.preattach_occupancy_frozen {"cells": 7506, "resolution_m": 0.015,
                                "revision": 1, "truncated": false}
```

Not plumbing — that is **7506 occupied cells** reduced to world keys at the real grasp, at real grid scale, with the revision matching the attach marker and **`truncated: false`**, so the 2 M cap is nowhere near binding. It also independently confirms `resolution_m: 0.015`, i.e. #269's default is what actually runs.

### 2. The annotation reaches a stop record on a live graph — proven

`tests/integration/test_estop_voxel_backing_live.py`, production `ManifestHALLifecycleNode`, real `openral_msgs` on the wire: cell in the frozen set → `preexisting: true`; one lattice step away → `false`; no snapshot → `available: false` with a reason.

### 3. Both in the same production run — **NOT achieved**

The host **OOM-killed the battery twice** (15 GB total; deploy graph + XR-1 sidecar). The first attempt died mid-round after the freeze, before any stop. The second died earlier still — the attach was *staged* but the kernel never acked it, so the applied path correctly never ran and no freeze happened. Worth stating plainly because it looks like a bug and is not: `_freeze_preattach_occupancy` lives in the applied branch, and that run never got there.

**So #272's actual question — does an `attached_payload` stop's cell predate the grasp — is still unanswered.** Nothing here should be read as answering it. What is established is that the instrument which will answer it works at both ends.

This box is the known-tight one (a whole-suite pytest OOMs here too). The battery harness's own header says it is designed for `spark`. The remaining run is a scheduling problem, not a code question.

### Also in this push

`1d21785` — the disconnect path cleared the frozen set but left `_preattach_stamp_ns` / `_preattach_revision` behind, so a later record would print provenance for a snapshot that no longer existed. Found while reviewing my own diff during the battery, fixed after it, both test tiers re-run green.

AdrianLlopart:
## Battery run, in full. Retracting the OOM claim — that was my fault.

### The OOM was mine, not the host's

I reported that this host could not hold a deploy round. Wrong. I had set `OPENRAL_SKIP_ORPHAN_REAP=1`, copied from `tools/ceiling_battery.sh`, where it exists because the battery's **parallel** workers cannot distinguish a concurrent sibling from a crash leftover. I was running a **single** worker, so nothing reaped the previous graph and each attempt started stacked on the last. No kernel OOM ever fired — `dmesg` and `journalctl` carry no OOM-killer entry; the task supervisor killed it on low system memory. With reaping left on, a round completes in **108 s**.

### The result — 4 rounds, `fridge` / gate-ON, on this branch

| round | kernel | certified true gap | cell verdict | classes |
| --- | ---: | ---: | --- | --- |
| r01 | −19.89 mm | **+0.00 mm** | **`attached_payload`** | `['attached_payload']` |
| r02 | −5.26 mm | +2.62 mm | `solid_world` | `['attached_payload', 'solid_world']` |
| r03 | −32.82 mm | +3.49 mm | `solid_world` | `['solid_world']` |
| r04 | — | — | no stop | **task succeeded** |

**r01 answers #272.** The carried payload stopped on a cell already occupied at the instant of its own grasp (`preexisting: true`, `cell_key [273, -50, 61]`, frozen set 7912 cells), whose only backing geometry is `obj_main`'s own `obj_g0` / `obj_g3`. Measured, not inferred.

### The caveat matters as much as the result

All three stops report `preexisting: true`, and **that number alone is worthless** — a world-backed cell is *always* pre-existing, because the fixture was there before the grasp. The payload-authored case is the **conjunction**: verdict `attached_payload` (no `solid_world` in the cell at all) *and* the cell predates the attach. **One of three, not three of three.**

I nearly reported the useless version. It is now written into `preattach_verdict`'s docstring and the `docs/methods/01-hal.md` entry so the field cannot be quoted the wrong way later.

### Two reading traps, both of which caught me first

1. The backing record lands on **either** the snapshot line or the evidence line, depending on which wins the race — 26 of 27 late in the 2026-09-09 battery, but r01 here was early. Grep both, or a present record reads as absent. (This is the same class of mistake that produced this issue's wrong original premise.)
2. The certified gap lives **only** on the snapshot line, so a stop whose backing arrived late has its backing and its ground truth on two different lines. They must be **joined on `stop_seq`**.

### Recorded, not chased

- **r04 is the first gate-ON `fridge` completion on record** (2026-09-09 was 0 of 5). n=4, Fisher `p ≈ 0.44` — an observation, **not** a rate claim, and nothing here says the gate's cost has moved.
- **r03 does not fit the geometry story.** 36.3 mm of excess on a `solid_world` cell, at 15 mm, with the tight hull already in. Neither #272 nor #266 box overhang. Unexplained at n=1.

### Verification

Unit 25 pass; full unit tier 5454 pass; live-ROS integration passes; `ruff`, `mypy --strict -p openral_hal`, `refresh_methods_linenos.py --check` all clean. Battery artifacts were host-local and are not committed.

